PR #1, opened 2015-02-20
Fix/little refactoring and active old disabled e2e test

vied12:
1. Before I work on the drafted post feature, I wanted to reorganize the `liveblog-edit` module.  
Roughly, I set the `$scope` with `angular.extend($scope, {...});`  
It helps me to understand what belongs to the scope or not.  
I also removed a `$watch`. I think it's good to avoid it when we can. See http://www.benlesh.com/2013/10/title.html
1. I renamed the button _NEW_ above the editor by _RESET_. I think this is what it does isn't it ?
1. I reactivated an e2e test which failed before. Now it works, maybe because of an update from superdesk, or because we are now using travis. Who knows ?


vladnicoara:
LGTM

PR #2, opened 2015-02-23
chore(server: requirements): use pypi versions of behave and wooper

actionless:


server/requirements.txt
```diff
@@ -35,10 +35,10 @@ six==1.7.3
...
 python3-ldap==0.9.5.3
+behave==1.2.5
+wooper==0.4.0
 
 git+git://github.com/superdesk/eve@upstream-nk#egg=Eve==0.6-dev
-git+git://github.com/actionless/behave@master#egg=behave
 git+git://github.com/amagdas/tinys3@list-buckets#egg=tinys3
-git+git://github.com/actionless/wooper@0.2.0#egg=wooper==0.2.0
 -e hg+http://bitbucket.org/haypo/hachoir3/#egg=hachoir3
 git+git://github.com/liveblog/superdesk.git#egg=Superdesk-Server
```
actionless: @vied12 u need to return a reference to the precise tag of superdesk here

actionless:
package name is correct(https://pypi.python.org/pypi/wooper/0.4.0), mb we need to wait for sorta reindex or smth

PR #3, opened 2015-02-23
Feature/auto updates on backend (LBSD 302)

vied12:


server/liveblog/blogs/blogs.py
```diff
@@ -34,13 +53,31 @@ class BlogsResource(Resource):
...
+        cid = update_key('blog_cid', flag=True)
+        if cid:
+            item['cid'] = cid
+        return cid
+
+
+def get_last_cid(item):
+    cid = update_key('blog_cid', flag=False)
+    if cid:
+        item['cid'] = cid
```
vied12: Here the test fails because `item` is None

server/liveblog/blogs/blogs.py
```diff
@@ -34,13 +53,31 @@ class BlogsResource(Resource):
...
         }
     }
 
+    item_methods = ['GET', 'PATCH', 'PUT', 'DELETE']
     privileges = {'GET': 'blogs', 'POST': 'blogs', 'PATCH': 'blogs', 'DELETE': 'blogs'}
 
 
-class BlogService(BaseService):
+def set_cid_on_blogs(item):
+        cid = update_key('blog_cid', flag=True)
```
vied12: Indentation is wrong

server/liveblog/blogs/blogs.py
```diff
@@ -51,7 +88,13 @@ def on_create(self, docs):
     def get(self, req, lookup):
         if req is None:
             req = ParsedRequest()
-        return self.backend.get('blogs', req=req, lookup=lookup)
+        docs = super().get(req, lookup)
+        return docs
+
+    def find_one(self, req, **lookup):
+        doc = super().find_one(req, **lookup)
+        get_last_cid(doc)
```
vied12: For a new liveblog developer like me it's hard to understand why we use the getter `get_last_cid` here without keeping the returned value.
It's also hard to guess the role and the meaning of `cid`. What does it mean ?
Maybe you should rename the function by replacing the `get_last_cid` by  `update_last_cid` ?

server/liveblog/blogs/blogs.py
```diff
@@ -34,13 +53,31 @@ class BlogsResource(Resource):
...
+            'type': 'integer'
         }
     }
 
+    item_methods = ['GET', 'PATCH', 'PUT', 'DELETE']
     privileges = {'GET': 'blogs', 'POST': 'blogs', 'PATCH': 'blogs', 'DELETE': 'blogs'}
 
 
-class BlogService(BaseService):
+def set_cid_on_blogs(item):
```
vied12: @ioanpocol said:

> One key is used for all blogs. To create a key for every blog you can use something like this: key = 'blog_cid' + str(item.blog)

server/liveblog/blogs/blogs.py
ancafarcas: The item can't be none and it is strange since for me none of the tests are failing

server/liveblog/blogs/blogs.py
ancafarcas: you're right.Will change it

server/liveblog/blogs/blogs.py
ancafarcas: cid is a counter which keeps track of my items/posts. It must be kept on blog level so that's why the method get_last_cid is called when one specific blog is requested.
For example: you have a blog, let's say BLOG1 and for this blog you want to create a POST. Your new created post has now cid=1. If you modify the post, the value of the cid will be 2. When you retrieve the BLOG1, among all the details you will also receive cid=2.

Was this description helpful?

server/liveblog/blogs/blogs.py
ancafarcas: Will change that. Thanks.

server/liveblog/blogs/blogs.py
vied12: Yes, better :)
So I think we can rename this function because it's not a getter, it also updates an attribute.
And what does the acronym `cid` means ?

server/liveblog/blogs/blogs.py
vied12: Do you run the test on the new repository ? Did you take a look on https://travis-ci.org/superdesk/liveblog/jobs/51839200 ?

ancafarcas: I have run only the behave tests. I will run also the e2e tests.

server/liveblog/blogs/blogs.py
ancafarcas: Ok. I'll give another name to the function. "cid" stands for change id :)

PR #5, opened 2015-02-24
chore(docker): add docker-compose plan and related scripts

actionless:


vied12:
Is there any way to remove every docker files from `server/` and `client/` and keep them only in `docker/` ?
In other words, having everything what is docker related in `docker/` ?

actionless:
No. On build docker client sends to docker daemon build context. Build context is a whole content of a directory where Dockerfile is stored.

nistormihai:
Looks goo I think

actionless:
squashed

vied12:
Reopened in #28

PR #9, opened 2015-02-26
feat: autoupdate with autobahn.

nistormihai:
LBSD-303

### DIFF
--- a/client/app/scripts/liveblog-edit/timeline.js
+++ b/client/app/scripts/liveblog-edit/timeline.js
@@ -10,6 +10,9 @@ define([
         var blog = {
             _id: $route.current.params._id
         };
+        $scope.$on('posts', function() {
+            console.log('posts');
+        });
         $scope.posts = [];
         $scope.totalPosts = 0;
         $scope.timelineLoading = false;
--- a/client/app/scripts/liveblog-edit/timeline.js
+++ b/client/app/scripts/liveblog-edit/timeline.js
@@ -1,7 +1,8 @@
 /*jshint nonew: false */
 define([
-    'angular'
-], function(angular) {
+    'angular',
+    'moment'
+], function(angular, moment) {
     'use strict';
     TimelineController.$inject = ['api', '$scope', '$rootScope', 'notify', 'gettext',
                                 '$route', '$q', '$cacheFactory', 'userList', 'itemsService'];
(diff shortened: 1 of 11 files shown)


server/requirements.txt
```diff
@@ -41,4 +41,4 @@ wooper==0.4.1
 git+git://github.com/superdesk/eve@upstream-nk#egg=Eve==0.6-dev
 git+git://github.com/amagdas/tinys3@list-buckets#egg=tinys3
 -e hg+http://bitbucket.org/haypo/hachoir3/#egg=hachoir3
-git+git://github.com/liveblog/superdesk.git@1eab0207a09f026c7875239da94a1d9a3199b14a#egg=Superdesk-Server
+git+git://github.com/liveblog/superdesk#egg=Superdesk-Server
```
vied12: It's important to keep a reference of the version here to keep the tests consistent.
@actionless gave me a lesson already :trollface:

client/app/scripts/liveblog-edit/timeline.js
```diff
@@ -10,6 +11,45 @@ define([
...
+            ];
+            return api('blogs/<regex(\"[a-f0-9]{24}\"):blog_id>/posts', blog).query({
+                source: {
+                    query: {filtered: {filter: {
+                        and: filter
+                    }}},
+                    sort: [{versioncreated: 'asc'}]
+                }
+            }).then(function(data) {
+                $scope.lastUpdated = moment.utc().format();
```
vied12: What happens if the user time is delaying ?
Why don't get the date from the latest post in the response ?

server/requirements.txt
nistormihai: I think this is due to merge, I don't think Anca has change this.

client/app/scripts/liveblog-edit/timeline.js
nistormihai: What do you mean by user time is delaying, it is using the utc  witch is the same for all :)
any way this will be reflector with the cid, that Anca has made.

vied12: Moment.js use the user clock time to compute the utc time, in this case we should set the date every time with the same source, the server time, IMO

nistormihai: Yes, so if your time is set off it can be set off aswell, but then this is your fault :),
Anyway like I sayed this is a temp solution the proper solution is with cid.

PR #10, opened 2015-02-26
LBSD-369 Improvements to delete on timeline

vladnicoara:
Item removal it's done by clicking a small 'x' sign
Posts are removable now by an action from the options drop-down

### DIFF
--- a/client/app/scripts/liveblog-edit/styles/liveblog-edit.less
+++ b/client/app/scripts/liveblog-edit/styles/liveblog-edit.less
@@ -511,8 +511,8 @@
         padding: 10px 0px 10px 0px;
         display: inline-block;
     }
-    .timeline-item-actions {
-        padding-top:25px;
+    .timeline-post-actions, .timeline-item-remove {
+        padding-top:9px;
         margin-right:10px;
         float: right;
         display: none;
--- a/client/app/scripts/liveblog-edit/timeline.js
+++ b/client/app/scripts/liveblog-edit/timeline.js
@@ -143,6 +143,18 @@ define([
         return service;
     }])
     .controller('TimelineController', TimelineController)
+    .directive('rollshow', [function() {
+        return {
+            link: function(scope, elem, attrs) {
+                elem.parent().on('mouseover', function() {
+                    elem.show();
+                });
+                elem.parent().on('mouseout', function() {
+                    elem.hide();
+                });
+            }
+        };
+    }])
     .directive('lbTimelineItem', ['api', 'notify', 'gettext', 'asset', 'itemsService', function(api, notify, gettext, asset, itemsService) {
         return {
             scope: {
@@ -173,7 +185,6 @@ define([
                         }
                     }
                 };
-
                 //only text items are collapsable
                 if (scope.post.type === 'text') {
                     if (scope.post.text.length > 200) {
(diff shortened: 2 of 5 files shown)


client/app/scripts/liveblog-edit/timeline.js
```diff
@@ -197,12 +197,42 @@ define([
...
     }])
-    .directive('itemActions', ['api', 'notify', 'gettext', 'asset', 'itemsService', 'modal',
+    .directive('postActions', ['api', 'notify', 'gettext', 'asset', 'itemsService', 'modal',
         function(api, notify, gettext, asset, itemsService, modal) {
         return {
             restrict: 'E',
             replace: true,
-            templateUrl: 'scripts/liveblog-edit/views/timeline-item-actions.html',
+            templateUrl: 'scripts/liveblog-edit/views/timeline-post-actions.html',
+            link: function(scope, elem, attrs) {
```
vied12: It would be maybe better to use an isolated scope for directives. It prevents to erase properties and make clear what is available in the scope.

It's not only me who tells it :)
From the [angular doc](https://docs.angularjs.org/guide/directive)
> Best Practice: Use the scope option to create isolate scopes when making components that you want to reuse throughout your app.

client/app/scripts/liveblog-edit/styles/liveblog-edit.less
```diff
@@ -511,12 +511,18 @@
...
     }
-    .timeline-item-actions {
-        padding-top:25px;
+    .timeline-item-remove {
+        padding-top:9px;
         margin-right:10px;
         float: right;
         display: none;
     }
+    .timeline-post-actions {
```
vied12: this is really similar to `.timeline-item-remove`
Can't we merge them ?

client/app/scripts/liveblog-edit/timeline.js
```diff
@@ -197,12 +197,42 @@ define([
             }
         };
     }])
-    .directive('itemActions', ['api', 'notify', 'gettext', 'asset', 'itemsService', 'modal',
+    .directive('postActions', ['api', 'notify', 'gettext', 'asset', 'itemsService', 'modal',
```
vied12: You never use `asset`

client/app/scripts/liveblog-edit/timeline.js
```diff
@@ -197,12 +197,42 @@ define([
...
+                        });
+                    });
+                    function confirm() {
+                        return modal.confirm(gettext('Are you sure you want to delete the post?'));
+                    }
+                };
+            }
+        };
+    }])
+.directive('itemRemove', ['api', 'notify', 'gettext', 'asset', 'itemsService', 'modal',
```
vied12: You never use `asset`

client/app/scripts/liveblog-edit/timeline.js
```diff
@@ -197,12 +197,42 @@ define([
...
+            templateUrl: 'scripts/liveblog-edit/views/timeline-post-actions.html',
+            link: function(scope, elem, attrs) {
+                scope.removePost = function() {
+                    var remItems = scope.post.groups[1].refs;
+                    confirm().then(function() {
+                        api.posts.remove(scope.post).then(function(message) {
+                            notify.pop();
+                            notify.info(gettext('Post removed'));
+                            scope.remove({post:scope.post});
+                            for (var i = 0, item; i < remItems.length; i ++) {
```
vied12: I don't think this is the role of the client to remove every items, it should be the server role.
You should only remove the post here IMO

client/app/scripts/liveblog-edit/styles/liveblog-edit.less
nistormihai: even if is similar we should keep classes for different styles.
maybe just have `display: none;` for `.timeline-item-remove` and define them in common.

client/app/scripts/liveblog-edit/timeline.js
nistormihai: @vied12 I don't agree with this, even if I don't like so much the `business logic` inside the client, if this will complicate a lot the server or even make more unreadable the server code we need to make it on the client.
also remember that is no transaction  on the server alswell so it can `fail` aswell on the same steps.

vied12: But here it complicate the client and increase (sometimes a lot!) the number of requests which can take some times.

Anyway this logic should go the `PostService` if you really want it in the client

PR #15, opened 2015-03-04
Timeline: refactored and uses PostService

vied12:
And repair the build grunt task

### DIFF
--- a/client/app/scripts/liveblog-edit/timeline.js
+++ b/client/app/scripts/liveblog-edit/timeline.js
@@ -4,50 +4,14 @@ define([
 ], function(angular) {
     'use strict';
     TimelineController.$inject = ['api', '$scope', '$rootScope', 'notify', 'gettext',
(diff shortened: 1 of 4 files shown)


client/app/scripts/liveblog-edit/timeline.js
```diff
@@ -4,50 +4,14 @@ define([
...
-                    // check if undefined in order to support old model without this field.
-                    // (before the Draft-Posts Feature)
-                    return typeof(post.post_status) === 'undefined' || post.post_status === 'open';
-                });
-                for (var j = 0, post = {}; j < posts.length; j ++) {
-                    //build the new post system
-                    post = posts[j];
+                                 $route, $q, $cacheFactory, userList, itemsService, postsService) {
+
+        function retrievePosts() {
```
nistormihai: can we make all the post processing inside the service?
here from my POV should only set the to scope.

vied12: Yes I would like to include the items manager in the post service

PR #16, opened 2015-03-09
Fix/refactor timeline with post service

vied12:
- Connect the post's `Edit` buttons in the timeline
- Simplify the timeline and remove the `itemService`
- Timeline style in a separated less file


vladnicoara:
:+1:

PR #19, opened 2015-03-11
bf_lb settings

ancafarcas:


ioanpocol:
:+1: 

Some e2e tests are failing. Probably frontend should be updated for blog them/language

PR #20, opened 2015-03-11
Feature/(draft posts) extend a post

vied12:
- posts in draft posts list use the same directive than in the timeline
- `lb-timeline-post` renamed to `lb-post`and accept a `allow-quick-edit` parameter
- Most of the directives declared in timeline.js are moved to directives.js

Contains https://github.com/superdesk/liveblog/pull/18


client/app/scripts/liveblog-edit/views/draft-posts-list.html
```diff
@@ -1,23 +1,11 @@
...
-            </div>
-            <article lb-bind-html html-content="{{ post.mainItem.item.text }}"></article>
-            <div  ng-pluralize class="draft-posts__items-number" ng-if="post.multiple_items" count="post.multiple_items - 1"
-                when="{
-                    'one': 'and one more item.',
-                    'other': 'and {} more items.'}">
-            </div>
-        </li>
+        <lb-post
+          ng-repeat="post in draftPosts.posts"
```
plamut: Does this post list get updated more than once? If so, it's strongly recommended to add "track by" in order to avoid unnecessarily re-creating DOM nodes.

vladnicoara:
LGTM

client/app/scripts/liveblog-edit/views/draft-posts-list.html
vied12: taken into account. Thanks ;)

plamut: I see this quite often (and not only in this project), it seems that most people are really not aware of the negative impacts on the performance, yet it's so easy to avoid them. Spread the word, please. :)

PR #21, opened 2015-03-12
Feature/auto updates

nistormihai:


client/app/scripts/liveblog-edit/posts.service.js
```diff
@@ -123,24 +230,21 @@ define([
...
-            return api.posts.remove(post).then(function() {
-                $rootScope.$broadcast('lb.posts.updated');
-                $rootScope.$broadcast('lb.posts.removed', post);
-            });
+            return api.posts.remove(post);
         }
 
         return {
+            posts: posts,
+            postsInfo: postsInfo,
```
vied12: `postInfo` is never used and if you use `$cacheFactory`, you can call the method `cache.info` to get the size if you need it.

nistormihai: in postInfo for now we keep the total number of posts,
which is needed by the pagination and this total number is compiled every time a post was added or removed. see updatePosts function.

client/app/scripts/liveblog-edit/posts.service.js
```diff
@@ -123,24 +230,21 @@ define([
...
         function removePost(post) {
-            return api.posts.remove(post).then(function() {
-                $rootScope.$broadcast('lb.posts.updated');
-                $rootScope.$broadcast('lb.posts.removed', post);
-            });
+            return api.posts.remove(post);
         }
 
         return {
+            posts: posts,
```
vied12: `posts` and `getPosts` are confusing, do we need both of them ? Can't we rename one ?

client/app/scripts/liveblog-edit/posts.service.js
```diff
@@ -47,12 +55,110 @@ define([
...
                 });
         }
+        function retrieveItems(blog_id, items_criteria) {
+            return api('blogs/<regex(\"[a-f0-9]{24}\"):blog_id>/items', {_id: blog_id})
+                .query(items_criteria)
+                .then(function(data) {
+                    return data._items;
+                });
+        }
+        function updateLastIndex(blog) {
```
vied12: Hard for me to understand the meaning of this function.
Why this function require a blog as parameter ?
What are the consequences if we call this function ?
It should be more explicit because this method is accessible from the service's API

client/app/scripts/liveblog-edit/posts.service.js
nistormihai: `posts` is the reference of the `getPosts` method, 
so actually the `posts` is just a link and `getPosts`  is the old method.
if is confusing ... I can rename to `requirePosts` or `fetchPosts`

vied12: `fetchPosts` sounds good

client/app/scripts/liveblog-edit/posts.service.js
vied12: Is `lastIndex` a list of last update dates ?
I would change the variable name with something more explicit.
Also I wonder why at the end of every methods that update posts or items we call `removeLastIndex();`.
Why should we remove the last update date when we update them ?
Shouldn't on the contrary add the new update date ?

PR #29, opened 2015-03-17
Fix/draft posts ui

vied12:
lbsd-403


client/app/scripts/liveblog-edit/styles/draft-posts.less
```diff
@@ -40,5 +41,133 @@
...
+            }
+            .post-item-content {
+                margin-left: 0px;
+                .post-item__header {
+                    font-size: 12px;
+                    .name {
+                        font-weight: 600;
+                    }
+                }
+                #timeline-dropdown.dropdown-menu {
```
vied12: @darconny Are you sure that all of that is necessary ?
We should at least rename all classes containing _timeline_ in the name. Because we are in `draft-posts.less`.
Also here we saw an id `#timeline-dropdown`. It shouldn't be an id but a class since it appears several time in the document.

darconny:
+1

PR #33, opened 2015-03-24
Feature/blog settings

vied12:


ioanpocol:
There is an issue on e2e tests:
PhantomJS 1.9.8 (Linux) dashboard can load user widgets FAILED

Error: [$injector:unpr] Unknown provider: $templateRequestProvider <- $templateRequest <- $route <- superdesk


:+1: (except above issue)

client/app/scripts/liveblog-edit/module.js
```diff
@@ -167,19 +167,52 @@ define([
...
+        // load available languages
         api('languages').query().then(function(data) {
-            vm.languages = data._items;
+            vm.availableLanguages = data._items;
+        });
+        // load available themes
+        api('themes').query().then(function(data) {
+            vm.availableThemes = data._items;
         });
+        // watch if the user selected preferences have changed, in order to update the `isSaved` variable
```
vladnicoara: I think that we should use a service for these api calls as I should use 3 of them in the exact same way (get for themes, languages and global_preferences)

vied12: We are already using a service, the `api` service.
This is the role of a controller to connect services to the view and this is what we do here.
We don't need another service to combine others services. IMO We will just have a more complex architecture.
But if you really want to respect the DRY principle, we should use a common controller for our respective settings, and if they diverge, use inheritance and extend them. But in my opinion, this is not necessary for these 3 simple methods.

server/requirements.txt
```diff
@@ -41,4 +41,4 @@ wooper==0.4.1
 git+git://github.com/superdesk/eve@upstream-nk#egg=Eve==0.6-dev
 git+git://github.com/amagdas/tinys3@list-buckets#egg=tinys3
 -e hg+http://bitbucket.org/haypo/hachoir3/#egg=hachoir3
-git+git://github.com/liveblog/superdesk@V0.1#egg=superdesk
\ No newline at end of file
+git+git://github.com/liveblog/superdesk@V0.1#egg=Superdesk-Server
```
nistormihai: why is this changed form #superdesk to #Superdesk-Server?

vied12: Because this is the name of the superdesk python dependency
https://github.com/superdesk/superdesk/blob/master/setup.py#L20

client/app/scripts/liveblog-edit/views/settings.html
```diff
@@ -36,23 +37,22 @@ <h2 translate>General</h2>
...
-                                    <select ng-model="settings.blogSettings.language" class="form-control">
-                                        <option>--- Default ---</option>
+                                    <select ng-model="settings.blogPreferences.language" class="form-control">
                                         <option
-                                          ng-repeat="language in settings.languages"
-                                          ng-selected="{{settings.blogSettings.language == language}}"
-                                          value="{{ language.language_code }}">{{ language.language_name }}</option>
+                                          ng-repeat="language in settings.availableLanguages"
+                                          ng-selected="{{ settings.blogPreferences.language == language.language_code }}"
+                                          value="{{ language.language_code }}">{{ language.name }}</option>
```
vladnicoara: it's better to use ng-options with ng-model if possible

client/app/scripts/liveblog-edit/module.js
```diff
@@ -167,19 +167,52 @@ define([
...
+                    vm.isSaved = true;
+                    vm.blog = blog;
+                    notify.pop();
+                    notify.info(gettext('blog settings saved'));
+                    vm.close();
+                });
+            },
+            reset: function() {
+                // reset vm.blogPreferences's values with the ones from global_preferences (backend)
+                api('global_preferences').query().then(function(global_preferences) {
```
nistormihai: `api` is a service but just a service to easeful work with api,
that doesn't exclude using other service to work across controller's functionality.
And using a common controller for something with services where build is not a good ideea.

ex: $http service and having a service that will fetch users, that is used in more than in one place.
 api('global_preferences') is used more then once.

nistormihai:
:+1:  beside the api.global_preferences

PR #34, opened 2015-03-25
Post actions simplified

vied12:
no more action to remove an item in a post and no more dropdown for them but a tooltip

LBSD-413

selection_048


vladnicoara:
LGTM

PR #39, opened 2015-03-30
Fix/Draft Post Panel width issue fixed

vied12:
And deep linking enabled (?drafts=open|close)

LBSD-417


nistormihai:
super, looks good to me.
We definitely need more permanent links to direct actions.

PR #45, opened 2015-04-02
Feature/general settings

vladnicoara:


client/app/scripts/liveblog-settings/styles/general-settings.less
```diff
@@ -0,0 +1,20 @@
+@PADDING: 20px;
+.general-settings {
```
vied12: It seems similar to `client/app/scripts/liveblog-edit/styles/settings.less`, we should factorize them.

client/app/scripts/liveblog-settings/module.js
```diff
@@ -0,0 +1,77 @@
+(function() {
+    'use strict';
+    LiveblogSettingsController.$inject = ['$scope', 'api', '$location', 'notify', 'gettext', '$q'];
+    function LiveblogSettingsController($scope, api, $location, notify, gettext, $q) {
+        //prep the settings
+        $scope.liveblogSettings = {'language': {}, 'theme': {}};
+        $scope.settingsLoading = true;
+        api.languages.query().then(function(data) {
+            $scope.languages = [];
```
vied12: I don't think you need to instantiate a new array here, You can assign `data._items` directly to `$scope.languages`.

client/app/scripts/liveblog-settings/module.js
```diff
@@ -0,0 +1,77 @@
...
+                $scope.languages.push(language);
+            });
+        });
+        api.themes.query().then(function(data) {
+            $scope.themes = [];
+            _.map(data._items, function(theme) {
+                $scope.themes.push(theme);
+            });
+        });
+        $scope.settingsLoading = true;
```
vied12: This line is already defined line 7.
Actually, if you were using `angular.extend($scope, {})` to define your scope, jsLint would told you that you have a duplicated definition of `settingsLoading`.

client/app/scripts/liveblog-settings/styles/general-settings.less
vladnicoara: I agree, when we'll refactor to get a common service, let's find a way to fix this too

PR #51, opened 2015-04-13
Fix/auto update refactored

vied12:
## Changelog
- `postService`  
  the post service has been very simplified (-150 lines if code)  
  His role has been limited to fetch the server and retrieve posts with a given pagination.
  The number of methods provided by the service has been also reduced.
- `lbPostsList` directive added  
  It is responsible of filling the view with a list of posts.  
  It is responsible of the next page selection  
  It also handle the auto update feature.
  Then its role is pretty well defined: managing the posts list.
- `draftPosts` directive removed. Replaced by `lbPostsList`.
- `TimelineController` controller removed. Replaced by `lbPostsList`.
- `posts/` endpoint: `deleted` is now a `boolean`
- e2e test added: "can edit an item on the timeline"


client/app/scripts/liveblog-edit/directives.js
```diff
@@ -50,40 +50,116 @@ define([
...
+                    var vm = this;
+
+                    function fetchPage() {
+                        // stop if the limit is reached
+                        if (angular.isDefined(vm.postsMeta.total) && vm.postsMeta.total <= vm.posts.length) {
+                            return false;
+                        }
+                        // Find the next page containing new posts
+                        var page = 1;
+                        while (vm.pagination.limit * page <= vm.posts.length) {
```
nistormihai: why while?

client/app/scripts/liveblog-edit/posts.service.js
```diff
@@ -22,6 +22,41 @@ define([
...
+                page: page,
+                max_results: max_results
+            };
+            // filters.status
+            if (angular.isDefined(filters.status)) {
+                posts_criteria.source.query.filtered.filter.and.push({term: {post_status: filters.status}});
+            }
+            // filters.updatedAfter
+            if (angular.isDefined(filters.updatedAfter)) {
+                posts_criteria.source.query.filtered.filter.and.push({
```
nistormihai: here `max_results` and `page` should be removed.

client/app/scripts/liveblog-edit/posts.service.js
```diff
@@ -44,204 +80,22 @@ define([
...
-                                angular.extend(post.items[indexItem].item, item);
-                            }
-                        }
-                    });
-                });
-            });
-        }
-
-        //pagination support for auto-update
-        function smartNextPage(blog_id, type, posts, postsInfo, max_results) {
```
nistormihai: why the pagination was removed?

client/app/scripts/liveblog-edit/directives.js
nistormihai: using normal pagination system for dynamical list will not work.
think if you are on page 2 and receive 3 pages of updates

vied12: For me, here this `while` is more explicit and natural than an euclidean division.
It's because I'm more a developer than a mathematician :), and given the low number of pages expected, it won't cost too much to the CPU.
Why it shouldn't work ? Can you explain the problem we will have ?
As it's done, for me, we will retrieve every time the next page containing new post. What ever the number of posts in the list.
For the updates, the mechanism is bellow.

client/app/scripts/liveblog-edit/posts.service.js
vied12: Can you tell why? I think it's good to have default values because it's a service, and service methods should be as simple as possible

client/app/scripts/liveblog-edit/posts.service.js
vied12: Pagination is now the `while` in the directive

nistormihai: this pagination was base on the `_created` date field now clasical pagination.

client/app/scripts/liveblog-edit/directives.js
nistormihai: updates mechanism is ok it is the old one, but the pagination you refactored it without being the purpose of this task or benign mention every where that it has changed. 
here is the example:
page one and two from a three page is shown, you receive 7 updates. page 3 is no composed of     7 posts from  page 2 posts and 7 from page 3.

vied12: I also changed the update mechanism, it is now totally separated from the pagination.
If I receive 7 updates, anyway, if I need a new page the while will find the next one containing new posts. Maybe it won't contains the maximum number of posts, but it's not an issue in an infinite scroll interface. Just get down to have more :)

client/app/scripts/liveblog-edit/posts.service.js
vied12: Yes and I guess it works because of the default order of `posts/`.

PR #56, opened 2015-04-16
Feature/general settings

vladnicoara:


nistormihai:
This is how it was supposed to be?
Also update it.

PR #58, opened 2015-04-17
timeline e2e test: wait for the update before checking the value of the text

vied12:
https://bamboo.sourcefabric.org/chain/result/viewChainResult.action?planKey=LBSD-DBRT&buildNumber=22#new-failed-tests maybe fixed.


nistormihai:
LGTM

PR #60, opened 2015-04-22
Fix/unpublish a post from the timeline (LBSD 11)

vied12:
- [x] Order draft posts list by update date
- [x] Don't update the reference of the post when saved as draft to don't disturb the `PostsListDirective` update mechanism, which is working with update date.
- [x] Active archive icon disappeared. Now fixed.


nistormihai:
LGTM tho wait for TRAVIS first.

PR #62, opened 2015-04-23
LBSD-38 Fix theme selection

vladnicoara:
It seems that for now when using ngOptions the select and label
expressions( as in select as label for value in array) can't be equal. Small hack until Angular fixes the problem.

### DIFF
--- a/client/app/scripts/liveblog-edit/views/settings.html
+++ b/client/app/scripts/liveblog-edit/views/settings.html
@@ -49,7 +49,7 @@
                                 <div class="form-input">
                                     <select
                                       ng-model="settings.blogPreferences.theme"
-                                      ng-options="theme as theme.name for theme in settings.availableThemes track by theme"
+                                      ng-options="theme.name as theme.name + ' ' for theme in settings.availableThemes"
                                       class="form-control">
                                     </select>
                                 </div>


client/app/scripts/liveblog-edit/views/settings.html
```diff
@@ -49,7 +49,7 @@ <h2 translate>General</h2>
                                 <div class="form-input">
                                     <select
                                       ng-model="settings.blogPreferences.theme"
-                                      ng-options="theme as theme.name for theme in settings.availableThemes track by theme"
+                                      ng-options="theme.name as theme.name + ' ' for theme in settings.availableThemes"
```
vied12: Why adding a trailing space ?
And why removing the `track by` ?

vladnicoara: Because in the way described by you, the whole theme object is sent to the model. You have to send only the theme name. You are using "label for value in array track by..". I am doing "select as label for value in array". I added the trailing space because for some reason when the select expression evaluates the same as the label expression the "selected" attribute doesn't work well on the option

vied12: Good to know! And what doesn't work well, what happens ?

vladnicoara: The set theme is not shown as selected when you refresh the page or return to the settings page. The selected="selected" attribute is there on the correct option but it does not show right in both FF and Chrome. Very strange behavior. I have tried with another field that I created, something like theme.displayName but had the same issue. Having the select and the value expressions evaluate to something just a bit different seems to fix it.

nistormihai:
LGTM

PR #63, opened 2015-04-24
Timeline small fixes (LBSD-11)

vied12:
- [x] Don't use the same post reference as in the postsList for `currentPost`
- [x] Timeline, open in the editor: check if item is defined before creating a block
- [x] Remove a post: don't let the service to update the reference of the post


nistormihai:
LGTM

PR #64, opened 2015-04-24
Feature/embed first page

nistormihai:
Final first page rendering with embed.
Access embed testing page to http://{{server}}/embed.html?liveblog.id={{blog_id}}


vied12:
LGTM

PR #65, opened 2015-04-27
LBSD-88 Show original creator in blog settings

vladnicoara:
Show image, full name and username

### DIFF
--- a/client/app/scripts/liveblog-edit/module.js
+++ b/client/app/scripts/liveblog-edit/module.js
@@ -194,8 +194,8 @@ define([
         });
     }
 
-    BlogSettingsController.$inject = ['$scope', 'blog', 'api', 'blogService', '$location', 'notify'];
-    function BlogSettingsController($scope, blog, api, blogService, $location, notify) {
+    BlogSettingsController.$inject = ['$scope', 'blog', 'api', 'blogService', '$location', 'notify', 'gettext'];
+    function BlogSettingsController($scope, blog, api, blogService, $location, notify, gettext) {
         // set view's model
         var vm = this;
         angular.extend(vm, {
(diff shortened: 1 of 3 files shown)


nistormihai:
LGTM

client/app/scripts/liveblog-edit/module.js
```diff
@@ -194,8 +194,8 @@ define([
         });
     }
 
-    BlogSettingsController.$inject = ['$scope', 'blog', 'api', 'blogService', '$location', 'notify'];
-    function BlogSettingsController($scope, blog, api, blogService, $location, notify) {
+    BlogSettingsController.$inject = ['$scope', 'blog', 'api', 'blogService', '$location', 'notify', 'gettext'];
+    function BlogSettingsController($scope, blog, api, blogService, $location, notify, gettext) {
```
vied12: Why importing `gettext` here ?

client/spec/blog_settings_spec.js
```diff
@@ -60,5 +60,24 @@ describe('Blog settings', function() {
...
+        openBlog(0);
+        openSettings();
+        element(by.css('[data="blog-settings-users"]')).click();
+        //expect original creator name
+        element(by.css('[data="original-creator-display-name"]')).getText().then(function(text) {
+            expect(text).toEqual('first name last name');
+        });
+    });
+
+    it('shows original creator username', function() {
```
vied12: It's maybe better to merge these two tests in the same in order to speed up the e2e tests

client/app/scripts/liveblog-edit/module.js
vladnicoara: it was used a bit later with notify.info(gettext('saving blog settings')); and travis was showing an error from time to time.... something like gettext module is not imported

client/spec/blog_settings_spec.js
vladnicoara: I thought of that as well but I separated them for ease of visibility when one of the expectations are not met

PR #69, opened 2015-04-30
Feature/liveblog list as default page

vied12:
- [x] default page set to `/liveblog/`
- [x] e2e tests updated
- [x] remove the workspace from the client
- [x] disable the karma tests from the client


nistormihai:
LGTM

PR #70, opened 2015-05-04
Feature/change blog owner

vladnicoara:


nistormihai:
:smile:

client/app/scripts/liveblog-edit/module.js
```diff
@@ -239,9 +242,25 @@ define([
...
+            });
+        }
+        buildOwner(blog.original_creator);
         // watch if the user selected preferences have changed, in order to update the `isSaved` variable
-        $scope.$watch(angular.bind(this, function () {return this.blogPreferences;}), function(new_value) {
-            vm.isSaved = _.isEqual(vm.blogPreferences, vm.blog.blog_preferences);
+        $scope.$watch(angular.bind(this, function () {return [this.blogPreferences, this.original_creator];}), function(new_value) {
+            vm.isSaved = _.isEqual(vm.blogPreferences, vm.blog.blog_preferences) &&
+            (!vm.original_creator._id || _.isEqual(vm.original_creator._id, vm.blog.original_creator));
+            //refresh owner data is owner changed
```
vied12: We must keep this function very simple, just to update `isSaved`. (Because $watch is very expensive)
If you need to follow the owner input, it's better to use ngChange

client/app/scripts/liveblog-edit/module.js
```diff
@@ -239,9 +242,25 @@ define([
         api('users').getById(blog.original_creator).then(function(data) {
             vm.original_creator = data;
         });
+        api('users').query().then(function(data) {
+            vm.avUsers = data._items;
```
vied12: I'm not sure this is the right way to retrieve the users. Think if we have more than one page of result.
I think we should use a field with auto completion making directly requests to the server

PR #73, opened 2015-05-06
Feature/show embed code

vladnicoara:


nistormihai:
LGTM

client/app/scripts/liveblog-edit/views/settings.html
```diff
@@ -100,3 +100,36 @@ <h2 translate>Users</h2>
         </div>
     </div>
 </div>
+<temnplate-holder id="liveblog-embed-template" type="text/template">
```
vied12: a typo here I guess, `template-holder` sounds better

client/app/scripts/liveblog-edit/module.js
```diff
@@ -243,6 +243,10 @@ define([
         $scope.$watch(angular.bind(this, function () {return this.blogPreferences;}), function(new_value) {
             vm.isSaved = _.isEqual(vm.blogPreferences, vm.blog.blog_preferences);
         }, true);
+        _.templateSettings.interpolate = /{{([\s\S]+?)}}/g;
+        var compiled = _.template(_.trim(document.getElementById('liveblog-embed-template').innerHTML));
+        /*globals config */
+        vm.embedCode = compiled({'rest': config.server.url, 'frontend': window.location.origin, id: vm.blog._id});
```
vied12: Why don't you use the basic template feature from angular ( `{{ }}` ) ?

client/app/scripts/liveblog-edit/module.js
```diff
@@ -243,6 +243,10 @@ define([
         $scope.$watch(angular.bind(this, function () {return this.blogPreferences;}), function(new_value) {
             vm.isSaved = _.isEqual(vm.blogPreferences, vm.blog.blog_preferences);
         }, true);
+        _.templateSettings.interpolate = /{{([\s\S]+?)}}/g;
+        var compiled = _.template(_.trim(document.getElementById('liveblog-embed-template').innerHTML));
```
vied12: You should select from the scope, not from the `document`

client/app/scripts/liveblog-edit/module.js
nistormihai: Why use the template from angular?
It is more clear the embed code, before and after the `minification` process.

client/app/scripts/liveblog-edit/module.js
nistormihai: that will just make the code more complex and will not be very clear.

vied12: My purpose was to isolate the query to prevent any issue with other view that can declare another `liveblog-embed-template` somewhere else.
Like you prefer. Then better to use `$document` or `angular.element()` with angular.

vied12: Actually, because you are manipulating element from the DOM, you should create a directive for that

nistormihai: Actually is not DOM is just a template to populate a textarea.

vied12: It's hidden in the DOM, but it's still a DOM element. Anyway, this kind of operation like `angular.element()` or `($)document.getElementById` shouldn't be done in the controller but in a directive if we want to follow the angular principles.

PR #76, opened 2015-05-11
LBSD-471 private draft posts

ancafarcas:


nistormihai:
LGTM

server/liveblog/posts/posts.py
```diff
@@ -129,8 +133,13 @@ def get(self, req, lookup):
         if lookup.get('blog_id'):
             lookup['blog'] = ObjectId(lookup['blog_id'])
             del lookup['blog_id']
+        logged_in_user = getattr(flask.g, 'user', None)
         docs = super().get(req, lookup)
         for doc in docs:
+            user_id = logged_in_user.get('_id')
+            creator_id = ObjectId(doc.get('original_creator'))
+            if creator_id == user_id and doc['post_status'] == 'draft':
+                doc['visible'] = True
```
vied12: We should exclude them from the response instead of setting a flag for privacy reason. Drafts must stay private and shouldn't be accessible from anyone else through the api.

PR #78, opened 2015-05-12
Feat/private drafts

ancafarcas:


server/liveblog/posts/posts.py
```diff
@@ -129,8 +133,13 @@ def get(self, req, lookup):
         if lookup.get('blog_id'):
             lookup['blog'] = ObjectId(lookup['blog_id'])
             del lookup['blog_id']
+        logged_in_user = getattr(flask.g, 'user', None)
         docs = super().get(req, lookup)
         for doc in docs:
+            user_id = logged_in_user.get('_id')
+            creator_id = ObjectId(doc.get('original_creator'))
+            if creator_id == user_id and doc['post_status'] == 'draft':
+                doc['visible'] = True
```
vied12: We should exclude them from the response instead of setting a flag for privacy reason. Drafts must stay private and shouldn't be accessible from anyone else through the api.

nistormihai: With this kind of implementation we can also request all the drafts, if it will be needed in the future.
@ancafarcas maybe it will be good to add this parameter as default to True for GET.

vied12: if it's needed in the future, it would be easy to add a parameter do retrieve them all.
What about privacy here ?
And what is the point to do it on the backend if it's only adding a flag ?

nistormihai: no it will not break the pagination caz the server can be interrogated only by this field.
this implementation is easier to make and cleaner to understand.

vied12: If you save this field "visible" on the database, you will have a surprise when you will login with another account :blush:

ancafarcas: As it is right now, if you create a draft post and fetch the posts you will get your post having "visible" set to true. If you log in with another account and fetch the posts, you will get the draft post created by the other user but it will have "visible" set to false

vied12: I know but it was to respond to @nistormihai . He supposed this field was saved to allow to request by this field.

ancafarcas: Yes, this was the idea. If you request by this field you can obtain all the draft posts of the logged in user (meaning the drafts which have "visible" set to true)

vied12: But how can you filter by a field that is not indexed in the database ?
It seems hard to do and very not efficient for the server's CPU

PR #93, opened 2015-05-22
Server: dependencies from SD removed from LB's requirements.txt

vied12:


nistormihai:
waw great job edouard

PR #111, opened 2015-06-04
Feature/Theme System (LBSD-511)

vied12:
https://gist.github.com/vied12/7c7163b21c12c3d95b80


server/liveblog/blogs/blogs.py
```diff
@@ -15,8 +15,9 @@
     'guid': metadata_schema['guid'],
     'title': metadata_schema['headline'],
     'description': metadata_schema['description'],
-    'language': Resource.rel('languages', True),
```
jey07ro: isn't language needed for a blog?

nistormihai: isn't language needed for a blog?

server/liveblog/blogs/blogs.py
```diff
@@ -15,8 +15,9 @@
...
-    'theme': Resource.rel('themes', True),
+    'theme': {
```
nistormihai: how so `dict` type? it should be a string IMO

server/liveblog/embed/embed.py
```diff
@@ -11,12 +11,28 @@
...
+bp = superdesk.Blueprint('embed_liveblog', __name__, template_folder='templates', static_folder=ASSETS_DIR)
 
 
 @bp.route('/embed/<blog_id>')
 def embed(blog_id):
-    return render_template('embed.html', blog_id=blog_id, api_host=request.url_root)
+
+    def complete_url(url):
+        def is_relative(url):
+            return not (url.startswith('/') or url.startswith('http://') or url.startswith('https://'))
```
nistormihai: starting with `/` maybe isn't a url at all it should start `//`

client/app/scripts/liveblog-edit/module.js
```diff
@@ -216,12 +216,13 @@ define([
...
-                        original_creator: vm.original_creator._id,
-                        blog_status: vm.blog_switch === true? 'open': 'closed'};
-                    angular.forEach(vm.newBlog, function(value, key) {
-                        changedBlog[key] = value;
-                    });
+                    blog_preferences: vm.blogPreferences,
+                    original_creator: vm.original_creator._id,
+                    blog_status: vm.blog_switch === true? 'open': 'closed'
+                };
+                angular.forEach(vm.newBlog, function(value, key) {
```
nistormihai: isn't this the same as _.create ?

server/liveblog/embed/embed_assets/default-theme/main.js
```diff
@@ -30,15 +30,17 @@
         vm.pagesManager.fetchNewPage()
         // retrieve updates periodically
         .then(function() {
-            $interval(retrieveUpdate, 2000);
+            $interval(retrieveUpdate, 10000);
```
nistormihai: muy bien!!!

server/liveblog/embed/embed_assets/default-theme/main.js
```diff
@@ -30,15 +30,17 @@
...
-            // change the template tag symbols to allow flask template tags
-            $interpolateProvider.startSymbol('[[').endSymbol(']]');
-        }])
-        .controller('timelineCtrl', TimelineCtrl);
+    angular.module('liveblog.default-theme', ['liveblog-embed', 'ngSanitize' ,'ngAnimate'])
+        .directive('lbTemplate', function() {
+            return {
+                controller: TimelineCtrl,
+                controllerAs: 'timeline',
+                templateUrl: window.LB_ASSETS_DIR+'/default-theme/default-theme-template.html'
```
nistormihai: isn't a good idea also to inject $window?

server/liveblog/embed/embed_assets/default-theme/package.json
```diff
@@ -0,0 +1,14 @@
...
+    "version": "1.0",
+    "angularModule": "liveblog.default-theme",
+    "styles": [
+        "styles/embed.css",
+        "//maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css"
+    ],
+    "scripts": [
+        "//code.angularjs.org/1.3.14/angular-sanitize.js",
+        "//code.angularjs.org/1.3.14/angular-animate.js",
+        "main.js"
```
nistormihai: can we have implied main.js?

server/liveblog/embed/templates/embed.html
```diff
@@ -1,64 +1,33 @@
 <!doctype html>
 <html ng-app="liveblog-embed">
     <head>
-        <title>Embed</title>
+        <title>{{ blog.title }}</title>
```
nistormihai: this is useless, we need to think on a plugin which will do some stuff for iframed window.

server/liveblog/blogs/blogs.py
vied12: `language` is set in `blog_preferences`

server/liveblog/blogs/blogs.py
vied12: in `blog_preferences`, theme is defined with a `string` (should be an `ObectId` maybe), but here `theme` is a snapshot of the theme, we copy all the theme definition to prevent any unwanted changes on this blog.

server/liveblog/embed/embed.py
vied12: This function is used to know if the path in the url is relative or not.
Is relative:
 - script.js
- script/main.js
Is not relative:
- /script/main.js
- //maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css
- https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css

Any way, I think we will need to exclude remote files (starting with the protocol or `//`) because we don't want to have to many `<script>` tags in the final embed html then we need to compile and minify them

client/app/scripts/liveblog-edit/module.js
vied12: I don't know what do you mean, I just corrected the indentation ^^

server/liveblog/embed/embed_assets/default-theme/main.js
vied12: sure it is, and even better: creating a `config` constant

server/liveblog/embed/templates/embed.html
vied12: It's useful for developers when you open the page in the browser, and for SEO.

You mean something like that http://npr.github.io/responsiveiframe/ ?

server/liveblog/embed/embed_assets/default-theme/main.js
vied12: :yum:

server/liveblog/embed/embed_assets/default-theme/package.json
vied12: Why? I don't see the point. Explicit isn't always better that implicit ?

client/app/scripts/liveblog-edit/module.js
nistormihai: underscore `_.create` is a magic, fast and awesome function.
But I see you don't use lodash or underscore yet, so maybe is better to use `angular.extend` instead.

server/liveblog/embed/embed_assets/default-theme/main.js
nistormihai: ok and then ...?

client/app/scripts/liveblog-edit/module.js
vied12: Well I understand. I fix it !
Btw it was your code ;) ccf5ce0f

server/liveblog/embed/embed_assets/default-theme/package.json
jey07ro: why not then add also angular?
Main the outside of the theme, so it shouldn't be in the theme.

server/liveblog/embed/templates/embed.html
nistormihai: yeah forgout about SEO.
not really what I had in mind with responsiveiframe but a good lib to use tho.
I was thinking more to interact with the embedable page, like adding in the title etc..

server/liveblog/embed/embed.py
nistormihai: when using cdn is good to have script tags in the embed, no matter how ugly it looks.
If we can choose, I will go with a hybrid solution: local files compiled and the other not.
local files, not relative. naming is relevant.

server/liveblog/embed/embed_assets/default-theme/package.json
nistormihai: why not then add also angular?
main.js the outside of the theme and from the scope of the theme, so it shouldn't be in the theme.

nistormihai: we could have before and after scripts if this is the reason.

server/liveblog/embed/templates/embed.html
nistormihai: ps. we don't deal with SEO at this point.

server/liveblog/blogs/blogs.py
nistormihai: interesting and complicated, also I don't see the point, if relevant theme files could still change.

vied12: This is to ensure that we freeze everything that is related to a blog when we publish it.
If someone change a theme file, it won't impact published blogs with the same theme.

server/liveblog/embed/embed.py
vied12: ok for the hybrid solution.
About naming, I insist, relative, not local. `/script/main.js` is not relative and not local.
(I'm talking about relative *path*, not relative *protocol*)

server/liveblog/embed/embed_assets/default-theme/package.json
vied12: Angular is already imported with the `liveblog-embed` module.
> main.js the outside of the theme and from the scope of the theme, so it shouldn't be in the theme

I don't understand sorry.
You prefer to have a `main.js` main with a forced name witch is always imported between two `before_script` and `after_script` templates instead of an explicit list of scripts ?

server/liveblog/embed/templates/embed.html
vied12: And then?? You want that I remove the title ? Sorry but having a page title is a must have and not only for SEO.

About the parent page, we can still use the API to retrieve data directly from the parent page.
But it's weird for a newspaper to update the page title through Liveblog. They have a CMS for that.

server/liveblog/embed/embed_assets/default-theme/package.json
nistormihai: in this version you added main.js into the theme. in the main it was outside of the theme.
ex:
https://github.com/superdesk/liveblog/blob/master/server/liveblog/embed/assets/scripts/liveblog-embed/main.js
vs
https://github.com/superdesk/liveblog/blob/feature/theme-system/server/liveblog/embed/embed_assets/default-theme/main.js

so everything is ok. Still I don't like that the main controller sits in the theme. but this discussion is over, pakage.json LGTM.

server/liveblog/embed/embed.py
nistormihai: "Absolute URLs are URLs that start with a scheme"

/script/main.js is relative.

http://www.coffeecup.com/help/articles/absolute-vs-relative-pathslinks/
http://en.wikipedia.org/wiki/HTTP_location

vied12: Ok then how should I name this function ?

nistormihai:
in rest of relative path method everything LGTM

server/liveblog/embed/embed.py
nistormihai: you can find a better name :)

PR #112, opened 2015-06-05
modal fix

darconny:
After update from SD, Create Live Blog modal has width issue. Fix provided here.


vied12:
The `form` that you are targeting belongs to `liveblog-bloglist`.
It's better to put your style here https://github.com/superdesk/liveblog/blob/master/client/app/scripts/liveblog-bloglist/styles/liveblog-bloglist.less

darconny:
@vied12 I have to disagree, as the form is part of the modal which can be used in any other page.

vied12:
Make sense :+1:

PR #114, opened 2015-06-09
Feature/Blog publishing on S3

vied12:
- [x] Allow to publish a blog (an html file) on S3
- [x] Publish the blog on S3 when it is created
- [x] Publish the blog on S3 when it is updated

Based on #111


nistormihai:
please update.
LGTM :star:

PR #115, opened 2015-06-11
Feature/show post creator in default theme LBSD-506

vied12:
selection_070


server/liveblog/client_modules/client_modules.py
```diff
@@ -95,4 +96,10 @@ def get(self, req, lookup):
         if req is None:
             req = ParsedRequest()
         docs = super().get(req, lookup)
+        # nest the user in the response
+        for doc in docs:
+            creator = get_resource_service('users').find_one(req=None, _id=doc.get('original_creator'))
+            # select fields that are useful
+            wanted_fields = ('first_name', 'last_name', 'display_name', 'username', 'picture_url')
+            doc['original_creator'] = {key: creator.get(key, None) for key in wanted_fields}
```
nistormihai: so you added the original_creator to the post lists.

nistormihai:
LGTM

server/liveblog/client_modules/client_modules.py
vied12: I did that to limit the number of requests, keeping in mind that we lost the ability to reset the cache if we update the user information...
If we see that the server support easily the traffic, let's remove that

PR #116, opened 2015-06-15
Fix/default theme

vied12:
- Fix the default theme
- Add an overview page showing a blog in all different themes
- In debug mode, use the local file system to retrieve themes


server/liveblog/embed/embed.py
```diff
@@ -28,14 +30,41 @@ def is_relative_to_current_folder(url):
...
 @bp.route('/embed/<blog_id>')
-def embed(blog_id, api_host=None):
+def embed(blog_id, api_host=None, theme=None):
     api_host = api_host or request.url_root
     blog = get_resource_service('client_blogs').find_one(req=None, _id=blog_id)
     if not blog:
         return 'blog not found', 404
-    theme_root = blog['theme']['name']
+    theme_name = theme
+    try:
```
nistormihai: I don't get all this part.
if theme is default None then why try catch? 
isn't this just simple as
`theme_name = theme or request.args.get('theme', None)`

nistormihai:
In rest LGTM.
please see travis is failing.

server/liveblog/embed/embed.py
vied12: Because `request` can return this error when this method is run outside from a request context.
This is the case when we publish the embed, this celery that calls this method.

PR #125, opened 2015-06-19
Fix/default theme

vied12:
- Issue with duplicated updated dates fixed
- Font changed from @darconny reviews
- General padding set to 10px
- Fix for an issue with twitter embed code (no url, embed code)


vladnicoara:
LGTM

PR #131, opened 2015-06-23
Feature/cache [LBSD-541]

vied12:
- Introduces a new class `BlogCache` to access the cache associated to a blog
- Uses Flask-Cache as interface to the cache system.  
  For now we are using SimpleCache, a simple memory cache from [werkzeug](http://werkzeug.pocoo.org/docs/0.10/contrib/cache/#werkzeug.contrib.cache.SimpleCache)
- Dissociates user information from posts to not make the cache more complex
- Remove server_ldap from travis tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ before_script:
 
 script:
     - if [ "${TARGET}" = "server" ]; then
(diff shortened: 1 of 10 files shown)


nistormihai:
LGTM

PR #133, opened 2015-06-24
Feature/show published date in timeline2

vladnicoara:


client/protractor-conf.js
```diff
@@ -2,7 +2,7 @@
 
 exports.config = {
     allScriptsTimeout: 30000,
-    baseUrl: 'http://localhost:9090',
```
vied12: Why ? it's working well with 9090

client/tasks/options/watch.js
```diff
@@ -10,18 +10,6 @@ module.exports = {
             '<%= appDir %>/scripts/liveblog-*/**/*.less'
         ]
     },
-    code: {
```
vied12: Again, why ? :)

vladnicoara: sorry, I commited some files that I use for local testing :baby: I will fix it

PR #138, opened 2015-06-25
Fix/settings cleaned

vied12:
This PR removes `user-roles` and `desks` tabs from settings and set the `liveblog` one as default when the user arrives in settings.


vladnicoara:
LGTM

PR #144, opened 2015-07-03
Feature/remove liveblog tab from settings

vladnicoara:


vied12:
LGTM but must be merged on `devel`, not ~~`master`~~

PR #146, opened 2015-07-06
Fix/settings content width

darconny:
- content width set to be equal for all tabs
- team edit modal position fixed

### DIFF
--- a/client/app/scripts/liveblog-edit/views/settings.html
+++ b/client/app/scripts/liveblog-edit/views/settings.html
@@ -40,7 +40,7 @@
                         <form class="form-horizontal" name="settings.forms.general">
                             <div class="form-group">
                                 <label for="inputTitle" class="control-label" translate>Blog Status</label>
-                                <div class="form-input top-pad">
+                                <div class="form-input">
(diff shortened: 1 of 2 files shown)


client/app/scripts/liveblog-edit/views/settings.html
```diff
@@ -202,7 +202,7 @@ <h3 class="inline-block" translate>Contributors</h3>
                                     <button class="close" ng-click="settings.cancelTeamEdit()"><i class="icon-close-small"></i></button>
                                     <h3 translate>Contributors</h3>
                                 </div>
-                                <div class="modal-body" style="height: 400px; max-height:400px;">
+                                <div class="modal-body">
                                     <div translate style="color: gray">ADD PEOPLE</div>
```
vied12: Would be better without any style in html

vied12:
Must be merged into __devel__, not ~~master~~

client/app/scripts/liveblog-edit/views/settings.html
darconny: Agree, thats why I removed it from html.

PR #147, opened 2015-07-06
Fix/settings content width

darconny:
- fixed content width on settings page
- modal position fix
- removed inline style


vied12:
:+1: :boar:

PR #150, opened 2015-07-07
Feature/theme system refactored [LBSD-551]

vied12:
The aim of this refactor is to allow developers to make theme without Angular.js.

It also introduces inheritance between theme.

Specs: https://gist.github.com/vied12/43dee2db26241073aa14


nistormihai:
LGTM

PR #154, opened 2015-07-09
Feature/common features in generic theme [LBSD-558]

vied12:
- `lb-twitter-card` directive created in *angular-base* theme
- `prettifyIsoDate` filter added to *angular-base* theme
- `update_blogs_themes` command added in order to update the themes from local files in database, set the new value to blogs and republish themes on s3 if --republish is given.


nistormihai:
Good decision with "lbTwitterCard" this is a good way to go with "plugin" system
LGTM

PR #157, opened 2015-07-15
feature: permalink initial draft.

nistormihai:
LBSD-61

### DIFF
--- a/server/liveblog/embed/embed_assets/themes/default-theme/main.js
+++ b/server/liveblog/embed/embed_assets/themes/default-theme/main.js
@@ -8,8 +8,8 @@
         var SHOW_AUTHOR = config.settings.showAuthor;
         var DEFAULT_ORDER = 'editorial'; // newest_first, oldest_first or editorial
         var UPDATE_EVERY = 10*1000; // retrieve update interval in millisecond
-        var PERMALINK_HASH = 'liveblog._id'; // the hash identifier for permalink.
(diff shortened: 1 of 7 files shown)


server/liveblog/embed/embed_assets/themes/default-theme/main.js
```diff
@@ -40,12 +53,58 @@
...
+                } else if(found) {
+                    $timeout(function(){
+                        $anchorScroll(vm.permalinkId);                        
+                    });
+                }
+            },
             pagesManager: new PagesManager(POSTS_PER_PAGE, DEFAULT_ORDER)
         });
+
+        vm.getPermalinkId();
```
vied12: A getter should return a value, and this value should be assigned when called, not inside the getter.

vied12:
I think it must available as an option in the theme settings. Because as you know it can break the hash parser of the parent page, or break the server request with unknown parameters (depending if it uses parameter or hash to share the post)

vied12:
Sorry I wasn't clear. I meant that the all feature should be an option because you still can break the parent page if it doesn't support unknown parameter.

Then we can introduce a new field to the options you already added:
```json
        {
            "name": "activeParmalink",
            "type": "boolean",
            "default": true
            ...
        },
        {
            "name": "permalinkHashMark",
            "dependsOn": [
                {"activeParmalink": true}
            ]
            ...
        },
        {
            "name": "permalinkHash",
            "dependsOn": [
                {"activeParmalink": true}
            ]
            ...
        }
```

PR #159, opened 2015-07-17
Feature/page objects

nistormihai:
the proposed option for chaining 

ex:

list
    .openBlog() // links to BlogPage object
    .openSettings() // links to BlogSettingsPage object
       .openTeam();

### DIFF
--- a/client/spec/helpers/pages.js
+++ b/client/spec/helpers/pages.js
@@ -1,0 +1,148 @@
+'use strict';
+
+var blogs = [
+    {title: 'title: end to end image', description: 'description: end to end image', username: 'first name last name'},
+    {title: 'title: end To end three', description: 'description: end to end three', username: 'first name last name'},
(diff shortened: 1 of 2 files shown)


client/spec/helpers/pages.js
```diff
@@ -0,0 +1,148 @@
...
+    maxLen = maxLen || 15;
+    var text = '';
+    var possible = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+    for (var i = 0; i < maxLen; i ++) {
+        text += possible.charAt(Math.floor(Math.random() * possible.length));
+    }
+    return text;
+}
+
+exports.list = new BlogListPage();
```
vied12: `list` from `pages` isn't really meaningful. A better name would explicit this is a a list of blogs. If it is :)

vied12:
Must be merged into devel instead of master

PR #166, opened 2015-07-27
LBSD-567 theme-removal

ancafarcas:


server/liveblog/themes/themes.py
```diff
@@ -81,6 +83,23 @@ def update_registered_theme_with_local_files(self):
                 created.append(theme)
         return (created, updated)
 
+    def on_delete(self, doc):
+        blogs_service = get_resource_service('blogs')
+        blogs = blogs_service.get(req=None, lookup=dict(blog_status='open'))
```
vied12: You should add in the lookup the removed theme to filter blogs, then you don't have to loop all over the blogs after to look after the ones that use the theme.

server/liveblog/themes/themes.py
```diff
@@ -81,6 +83,21 @@ def update_registered_theme_with_local_files(self):
                 created.append(theme)
         return (created, updated)
 
+    def on_delete(self, deleted_theme):
+        blogs_service = get_resource_service('blogs')
+        blogs = blogs_service.get(req=None, lookup={'theme._id': deleted_theme['_id']})
+        for blog in blogs:
+            # will assign the default theme to this blog
+            theme = blogs_service.get_theme_snapshot(blog['blog_preferences']['theme'])
```
vied12: This is not the default theme. The default theme can be found in the general settings.

server/liveblog/themes/themes.py
```diff
@@ -81,6 +83,21 @@ def update_registered_theme_with_local_files(self):
                 created.append(theme)
         return (created, updated)
 
+    def on_delete(self, deleted_theme):
+        blogs_service = get_resource_service('blogs')
```
vied12: First, you should check if it's not the default theme from the general settings. (Don't confuse "default-theme" which is a theme name, and "default theme" which refers to the theme select as default theme in the general settings)

server/liveblog/themes/themes.py
```diff
@@ -81,6 +83,21 @@ def update_registered_theme_with_local_files(self):
...
+        blogs = blogs_service.get(req=None, lookup={'theme._id': deleted_theme['_id']})
+        for blog in blogs:
+            # will assign the default theme to this blog
+            theme = blogs_service.get_theme_snapshot(blog['blog_preferences']['theme'])
+            try:
+                blogs_service.system_update(ObjectId(blog['_id']), {'theme': theme}, blog)
```
vied12: To update the theme, change the value of `blog_preferences.theme`. The `theme` property will be updated in `on_update()`

server/liveblog/themes/themes.py
ancafarcas: yep, I started to work on this issue since I realized the exact thing you are talking about

server/liveblog/themes/themes.py
ancafarcas: :+1:

server/liveblog/themes/themes.py
ancafarcas: :+1:

PR #169, opened 2015-08-07
LBSD-579 add a theme from theme manager

ancafarcas:


server/liveblog/themes/themes.py
```diff
@@ -70,9 +75,27 @@ def get_local_themes_packages(self):
             for file in glob.glob(embed_folder + '/**/theme.json'):
                 yield json.loads(open(file).read())
 
+    def get_themes(self):
+        res = get_resource_service('themes').get(req=None, lookup={})
```
vied12: you already are in the theme service, use `self` to retrieve it instead of `get_resource_service('themes')`

server/liveblog/themes/themes.py
```diff
@@ -70,9 +75,27 @@ def get_local_themes_packages(self):
...
+        return res
+
     def update_registered_theme_with_local_files(self):
```
vied12: I don't think this the best place to extract zip files.
You should extract it when the user uploads it.
Also, it's not necessary to extract it  in the local file system. In the media storage would be better because we use S3.

server/liveblog/themes/themes.py
```diff
@@ -25,6 +27,9 @@ class ThemesResource(Resource):
             'type': 'string',
             'unique': True
         },
+        'zippedfiles': {
```
vied12: Is there a reason to keep this file associated in the database ?
Can't we just extract it and remove it ?

server/liveblog/themes/themes.py
```diff
@@ -70,9 +75,27 @@ def get_local_themes_packages(self):
             for file in glob.glob(embed_folder + '/**/theme.json'):
                 yield json.loads(open(file).read())
 
+    def get_themes(self):
+        res = get_resource_service('themes').get(req=None, lookup={})
+        return res
+
     def update_registered_theme_with_local_files(self):
+        embed_folder = os.path.join(os.path.dirname(os.path.realpath(__file__)),
```
vied12: You probably already know the _"Don't repeat yourself"_ principle.  Here you can apply the _"Don't repeat myself"_ principle :)
This line already exist a little above

server/liveblog/themes/themes.py
ancafarcas: About the place to extract...I was thinking the same but it is specified in the description of the story that the launch of the action should be made here

server/liveblog/themes/themes.py
ancafarcas: yes, it exists but I need to use the embed_folder inside this method too. Any idea on how I can re-use the same embed_folder which is already defined?

vied12: you can put the path in a common variable. But it won't be necessary since, as I told you, you should save the files in the media storage

ancafarcas: :+1:

Issue #168, opened 2015-08-04
docker-local-demo.sh generates yml from a deleted script


doctorlard:
The yml file is now in the source tree as of commit 8d21921 and does not need to be created with a script.

PR #171, opened 2015-08-17
This fixes issue #168

doctorlard:
Missing script called from scripts/docker-local-demo.sh since docker-compose.yml is now in the source tree.


doctorlard:
(Gah! wrong refspecs)

PR #179, opened 2015-08-19
Feature/remove theme

ancafarcas:
based on #170 

## TODO

- [ ] e2e tests
- [x] check in the backend if the removed theme doesn't belong to the default theme dependencies


client/app/scripts/liveblog-themes/module.js
```diff
@@ -0,0 +1,161 @@
...
+                notify.pop();
+                notify.info(gettext('Default theme saved'));
+                $scope.globalTheme = data;
+            });
+
+            }
+        };
+
+        $scope.removeTheme = function(theme) {
+            api.themes.query().then(function(data) {
```
vied12: Like in the backend, you must never retrieve all the resources if you don't need them.
First because it will overload the server and kill it :angel:, and also because there is pagination that limits the number of responses.
Then you should use the `getById()` method instead of `query()`. Or continue to use `query`, but with a parameter to filter the query.

vied12:
merged into #187

PR #183, opened 2015-08-25
Add 15px between posts in timeline [LBSD-610]

vied12:
and Make images max width to 100% - timeline and drafts panel [LBSD-609]


vladnicoara:
LGTM

PR #184, opened 2015-08-25
Feature/allow to add two images in a post [LBSD-614]

vied12:
Remove a duplicated id for file upload


vladnicoara:
I don't see anything wrong here :)

PR #185, opened 2015-08-26
LBSD-586 misbehave on draft remove

ancafarcas:


vied12:
LGTM if you change the previous message here https://github.com/superdesk/liveblog/blob/fix/misbehave-draft-remove/client/app/scripts/liveblog-edit/directives.js#L197

You can remove it or change the message by "Removing the post..."

vied12:
:+1:

PR #186, opened 2015-08-26
Fix/only owner can change blog settings

vladnicoara:


client/app/scripts/liveblog-edit/module.js
```diff
@@ -173,6 +172,12 @@ define([
                     return true;
                 }
             }
+
+        });
+        blogSecurityService.isUserOwner().then(function(result) {
+            $scope.blog.showSettingsLink = result;
+        }, function() {
+            $scope.blog.showSettingsLink = false;
```
vied12: To avoid this kind of instructions lost in the controller, I would set it into the template directly where you need it.
```html
ng-if="blogSecurityService.isUserOwner(blog)"
```

As you can see, I would also add a `blog` parameter, because I think this function should be usable outside from the route context.

client/app/scripts/liveblog-edit/module.js
```diff
@@ -444,7 +487,12 @@ define([
             controller: BlogSettingsController,
             controllerAs: 'settings',
             templateUrl: 'scripts/liveblog-edit/views/settings.html',
-            resolve: {blog: BlogResolver}
+            resolve: {
+                blog: BlogResolver,
+                security: ['blogSecurityService', function(blogSecurityService) {
```
vied12: You don't use it? Then you should maybe remove it

client/app/scripts/liveblog-edit/module.js
```diff
@@ -430,6 +435,44 @@ define([
...
+                    def.resolve(true);
+                }
+            }, function(response) {
+                if (response.status === 404) {
+                    def.reject('Blog does not exist');
+                }
+            });
+            return def.promise;
+        }
+        function canGoToSettings() {
```
vied12: Methods like __can__Something, or __is__Something, shouldn't do any other action than returning a boolean. Here you redirect users to other route.

vied12:
:+1:

PR #187, opened 2015-08-26
Remove, upload and add a theme

vied12:
- [x] Allow to upload a zip file containing a theme
- [x] Themes assets moved to `liveblog/themes/themes_assets/`
- [x] Hide abstract theme from the available themes list
- [x] End to end tests for uploading a theme
- [x] Remove a theme
- [x] End to end tests for removing a theme
- [x] Replace tinys3 by mediastorage (boto)
- [x] Fix the s3 issue that return an error for resources and fix the content type for JavaScript files


client/app/scripts/liveblog-themes/module.js
```diff
@@ -193,6 +154,66 @@
...
+
+        function loadThemes() {
+            // load only global preference for themes.
+            return api.global_preferences.query({'where': {'key': 'theme'}}).then(function(global_preferences) {
+                $scope.globalTheme = _.find(global_preferences._items, function(item) {
+                    return item.key === 'theme';
+                });
+                // load all the themes.
+                // TODO: Pagination
+                return api.themes.query({timestamp: Date()}).then(function(data) {
```
plamut: It probably makes sense to flatten this promise chain.

server/liveblog/themes/themes.py
```diff
@@ -77,43 +96,127 @@ class ThemesResource(Resource):
...
+        # Determine if there is a root folder
+        roots = set(file.split('/')[0] for file in files)
+        root_folder = len(roots) == 1 and '%s/' % roots.pop() or ''
+        # Check if the package is correct
+        try:
+            description_file = next((file for file in files if file.endswith('theme.json')))
+            # decode and load as a json file
+            description_file = json.loads(zip_file.read(description_file).decode('utf-8'))
+        except StopIteration:
+            return json.dumps(dict(error='A theme needs a theme.json file')), 400
```
plamut: Instead of catching the `StopIteration` error, you can also simply pass a default value argument to `next()`:

```py
item = next(<iterable>, None)
if item is None:
    # return error

# do something with item...
```

vied12: But I don't want a default value, I expect there is an item.
Dealing with exception in this case is more pythonic IMO

plamut: No worries, I just mentioned it because my opinion about what is more Pythonic is just the opposite here. :) 
The `StopIteration` error doesn't catch the semantics that well IMO, the `if item is None` construct describes the intent in a more direct way. After all we are not interested in the loop(ing) itself, just in the (non-)existence of a single item.

client/app/scripts/liveblog-themes/module.js
```diff
@@ -75,11 +75,6 @@
                     'url': authorArray[3]
                 };
             }
-            // If the screenshot string is not a url then compose it from server api url.
-            var protocolRegex = /^(http[s]?:)?\/{2}/;
```
nistormihai: why absolute urls are not an options anymore? for screenshoot.

client/app/scripts/liveblog-themes/module.js
```diff
@@ -193,6 +158,66 @@
...
+                angular.extend($scope, {
+                    themesHierachy: themes_hierachy,
+                    themes: themes,
+                    loading: false
+                });
+                return themes;
+            });
+        }
+
+        $scope.uploadZippedTheme = function(e) {
```
nistormihai: the name is not relevant. it should be plain upload Theme
It is really up to the endpoint to establish the compression.

server/liveblog/themes/themes.py
```diff
@@ -77,43 +96,126 @@ class ThemesResource(Resource):
...
+                    content_type = mime.from_file(name).decode('utf8')
+                    if content_type == 'text/plain' and name.endswith(tuple(CONTENT_TYPES.keys())):
+                        content_type = CONTENT_TYPES[os.path.splitext(name)[1]]
+                    final_file_name = os.path.relpath(name, CURRENT_DIRECTORY)
+                    # remove existing first
+                    app.media.delete(final_file_name)
+                    # upload
+                    file_id = app.media.put(file.read(), filename=final_file_name, content_type=content_type)
+                    # save the screenshot url
+                    if name.endswith('screenshot.png'):
```
nistormihai: so basicly this is hardcoded?
we need to document this properly.

vied12: Yes, as wordpress does, we expect that themes contain a preview in a file named `screenshot.png`.

The theme structure and the screenshot dimensions will be defined into the documentation that explains how to create a theme.

nistormihai:
LGTM but a python review should be made.

vied12:
The python review was already made by @plamut, isn't it ?

plamut:
@vied12 Ugh, it was, but it was just a quick glance out of boredom, not a thorough one. :) Though Anca did a review, too?

ancafarcas:
Yes, Peter, I did some review for the back-end code :)

PR #188, opened 2015-08-27
LBSD-593 User should only be able to publish an article one time

vladnicoara:
When publishing a post, the "PUBLISH" button should get disabled after
the user clicks on it for the first time. Currently, the user can
publish the same post multiple times to the timeline by repeatedly
clicking on the "PUBLISH" button while the first request is still
processing.


vied12:
:+1:

PR #190, opened 2015-09-02
fix: blog image remove and e2e for this.

nistormihai:
e2e for expectBlog fix for blog without and with image.
getAttribute returns a promise, wrong approach for e2e blog image remove. always passing e2e.
blog `picture` and `picture_url` accepts null ( so we can remove the damn image via patch).


vied12:
:ok_hand:

PR #191, opened 2015-09-02
LBSD-622 Preview a Post

vladnicoara:


vied12:
Let's go with that for the moment, but we will need later to render it as it is in the timeline, and more generally, to factorize the post's styles.

nistormihai:
@vied12 preview like is in timeline FMPOV is useless, it should be previewed as it is in embed.

vied12:
The post rendered by the timeline is more similar to the embed than the
current implementation.
To preview as it is in the embed is not so easy because it depends of the
selected theme.
But it would be doable, by rendering the embed with only the previewed post
as data.

PR #194, opened 2015-09-03
LBSD-628 User is able to save the same draft multiple times

vladnicoara:


vied12:
More generally, this issue can occur with all the other actions, isn't it ?
Then it would be maybe more simple to disable all the actions when an other is performing.
@nistormihai @vladnicoara ?

nistormihai:
@vied12 yeah I think  so, but the real issue is on creating.
remove and de

nistormihai:
ps for now this looks good to mege for me.

vied12:
Why not having `operationOnGoing` instead of `saveAsDraftDisabled`, that lock all the actions when another one is performing ?
etag prevents the server to make mistake, but will not prevent the view to react strangely for the user

PR #197, opened 2015-09-07
feature/remove blog

ancafarcas:


server/liveblog/blogs/blogs.py
```diff
@@ -83,21 +83,13 @@ class BlogsVersionsResource(ArchiveVersionsResource):
...
-class BlogsVersionsService(BaseService):
-    def get(self, req, lookup):
-        if req is None:
-            req = ParsedRequest()
-        return self.backend.get('archive_versions', req=req, lookup=lookup)
-
-
 class BlogsResource(ArchiveResource):
     datasource = {
-        'source': 'archive',
```
vied12: How does it work for the source name when you don't explicit it ?
Isn't better to add `source: 'blogs'` like you did bellow ?

server/liveblog/blogs/__init__.py
```diff
@@ -15,7 +14,7 @@ def init_app(app):
     UserBlogsResource(endpoint_name, app=app, service=service)
 
     endpoint_name = 'blogs_versions'
```
vied12: Do we still need this `blogs_versions` endpoint ?

server/liveblog/blogs/blogs.py
```diff
@@ -196,11 +188,14 @@ def find_one(self, req, **lookup):
...
+    def is_owner(self, doc):
+        if not is_admin(get_user()) and str(flask.g.user['_id']) != str(doc['original_creator']):
+            raise SuperdeskApiError.forbiddenError(message='You need to be the blog owner to perform updates on it')
+
     def on_update(self, updates, original):
         # check permission (see https://github.com/superdesk/liveblog/pull/167)
         # only the owner can change blog's settings
-        if not is_admin(get_user()) and str(flask.g.user['_id']) != str(original['original_creator']):
-            raise SuperdeskApiError.forbiddenError(message='You need to be the blog owner to perform updates on it')
+        self.is_owner(original)
```
vied12: now every editor, owner and admin can update the blog settings

server/liveblog/blogs/blogs.py
```diff
@@ -83,21 +83,13 @@ class BlogsVersionsResource(ArchiveVersionsResource):
...
-        if req is None:
-            req = ParsedRequest()
-        return self.backend.get('archive_versions', req=req, lookup=lookup)
-
-
 class BlogsResource(ArchiveResource):
     datasource = {
-        'source': 'archive',
+        'search_backend': 'elastic',
         'elastic_filter': {'term': {'particular_type': 'blog'}},
```
vied12: is this `particular_type` field related to archive ? then I think that we don't need it any more

vied12:
@hlmnrmr @ancafarcas Isn't the right time to convert all our collections to classic Resources ? Posts and Items are still Archives

server/liveblog/blogs/blogs.py
ancafarcas: ok, I will check for the roles compatibility with changing blogs settings

server/liveblog/blogs/blogs.py
ancafarcas: yes, it is related to archive and it will be removed now

server/liveblog/blogs/blogs.py
ancafarcas: Either you let 'search_backend': 'elastic' or 'source': 'blogs' it will use as source the "blogs" collections. In order to keep things homogeneous we can keep   'source': 'blogs'.

vied12:
LGTM, finally :sweat_smile:

PR #200, opened 2015-09-10
fix: replace facebook embed code by permalink.

nistormihai:
LBSD-646


vied12:
What about adding this modification to https://github.com/superdesk/angular-embed/blob/master/src/angular-embed/handlers/facebook.js instead ?

Also I would recommend to always attach a comment to a regex explaining what it does.  I realize this is not easy to read when you didn't write it :)

nistormihai:
@vied12 that service is triggered for facebook url only, this is converting embed to url.
 will add the comment, ps: I thought it was straight forward, catch facebook url, kind a of like the regex in the service link you gave.

vied12:
LGTM

PR #201, opened 2015-09-10
LBSD-627 Administrator should be able to edit blog settings

vladnicoara:
Administrator should be able to edit blog settings even if not blog owner

### DIFF
--- a/client/app/scripts/liveblog-edit/module.js
+++ b/client/app/scripts/liveblog-edit/module.js
@@ -462,10 +462,10 @@ define([
             return privileges.userHasPrivileges({'publish_post': 1});
         }
         function isUserOwner(archive) {
-            if ($rootScope.currentUser._id !== archive.original_creator) {
-                return false;
-            } else {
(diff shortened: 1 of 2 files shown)


client/app/scripts/liveblog-edit/module.js
```diff
@@ -462,10 +462,10 @@ define([
             return privileges.userHasPrivileges({'publish_post': 1});
         }
         function isUserOwner(archive) {
-            if ($rootScope.currentUser._id !== archive.original_creator) {
-                return false;
-            } else {
+            if ($rootScope.currentUser._id === archive.original_creator || $rootScope.currentUser.user_type === 'administrator') {
```
vied12: @vladnicoara You created a `SecurityService` somewhere, isn't it ? :)
What about using it to check the permission ?

vladnicoara: This is the SecurityService, I am modifying one of the methods there

vied12: Oh yes it is... :yum: sorry
:+1:

vied12:
LGTM

PR #202, opened 2015-09-11
LBSD-634 Disable access to blog settings for a contributor even if owner

vladnicoara:


client/app/scripts/liveblog-edit/module.js
```diff
@@ -461,8 +461,14 @@ define([
...
         }
-        function isUserOwner(archive) {
-            if ($rootScope.currentUser._id === archive.original_creator || $rootScope.currentUser.user_type === 'administrator') {
+        function accessToSettings(archive) {
+            //if no role assigned, assume no restriction from the role POV
+            var rolePermission = true;
+            if ($rootScope.currentUser.role) {
+                rolePermission = rolePermission && privileges.userHasPrivileges({'blogs': 1});
+            }
+            if (($rootScope.currentUser._id === archive.original_creator && rolePermission) ||
```
vied12: An administrator has by default all the privileges. You don't have to check if he is administrator if you already check the privileges

vladnicoara: Editors also have the 'blogs' privilege but they only should have access to the settings of their own blogs while an administrator should be able to access the settings of all blogs

vied12: I repeat what we said together for the record :fax:

When no role are set, the user must not have the permission to edit settings.

client/app/scripts/liveblog-edit/module.js
```diff
@@ -461,8 +461,14 @@ define([
         function canPublishAPost(blog) {
             return privileges.userHasPrivileges({'publish_post': 1});
         }
-        function isUserOwner(archive) {
-            if ($rootScope.currentUser._id === archive.original_creator || $rootScope.currentUser.user_type === 'administrator') {
+        function accessToSettings(archive) {
+            //if no role assigned, assume no restriction from the role POV
+            var rolePermission = true;
+            if ($rootScope.currentUser.role) {
+                rolePermission = rolePermission && privileges.userHasPrivileges({'blogs': 1});
```
vied12: `rolePermission` is set to true just before. No need to test it.

vied12:
:+1:

PR #203, opened 2015-09-11
Feature/filter by user in contributions

vied12:


client/spec/helpers/pages.js
```diff
@@ -260,24 +260,40 @@ function AbstractPanelPage(blog) {
...
+        });
+    };
 }
-ContributionsPage.prototype = Object.create(AbstractPanelPage.prototype);
 
 function TimelinePage(blog) {
     var self = this;
     self.blog = blog;
-    self.column = element(by.css('.column-timeline'));
+    self.column = element(by.css('.column--timeline'));
```
nistormihai: why?

vied12: because it's like that in the DOM.
I refactored a little bit the class names here to add the contributions column.
And as I started a long time ago, I use the BEM syntax for css.

client/app/scripts/liveblog-edit/directives.js
```diff
@@ -145,6 +145,11 @@ define([
                             });
                         });
                     });
+                    $scope.$watch('lbPostsFilterAuthors', function(users) {
```
nistormihai: can we do it without a watch?

vied12: I will find a way, maybe with a callback

client/spec/helpers/pages.js
nistormihai: I thought it was implied that why is that in the DOM,

from what I see `--` is an alternative in BEM description, so why?

nistormihai:
LGTM

client/spec/helpers/pages.js
vied12: `column` is the base model, `column--timeline` is the special rule for this column.
`column--timeline` represents different state of `column`
http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/ is a good introduction to BEM. It can looks verbose at the first try, but BEM helped me a lot to structure my css and to order it.

client/app/scripts/liveblog-edit/views/filter-by-member.html
```diff
@@ -0,0 +1,41 @@
...
+<div class="dropdown-content" ng-click="$event.stopPropagation();" ng-if="vm.openSelector">
+    <div class="user-select">
+        <div>
+            <form>
+                <label for="assignee-search" translate>Select users</label>
+                <input type="text" name="assignee-search" ng-model="vm.search">
+            </form>
+            <div class="search-result">
+                <ul class="users-list-embed">
+                    <li ng-repeat="user in vm.members | filter:vm.search | orderBy: 'display_name'"
```
plamut: It is strongly recommended to use `track by` for collections that tend to change to avoid unnecessary DOM operations, and search results are a good candidate for that in most cases.

client/app/scripts/liveblog-edit/views/filter-by-member.html
```diff
@@ -0,0 +1,41 @@
...
+                        ng-click="vm.toggleUserInPreselection(user)"
+                        ng-class="{'active': vm.isUserInPreselection(user)}"
+                        data-username="{{user.username}}">
+                        <figure class="avatar small"><img sd-user-avatar data-src="user.picture_url"></figure>
+                        <div class="name">{{user.display_name}}</div>
```
plamut: Can this change in the scope's lifetime? If not, it is recommended to use one-time binding to reduce the total number of $watches.

PR #205, opened 2015-09-14
LBSD-643 Enhance blog list link

vladnicoara:
The link to the list of blogs in theme manager must be bigger


vied12:
Maybe you can still use one way binding here
/poke @plamut :ghost:

nistormihai:
LGTM

client/protractor-conf.js
```diff
@@ -2,7 +2,7 @@
 
 exports.config = {
     allScriptsTimeout: 30000,
-    baseUrl: 'http://localhost:9090',
+    baseUrl: 'http://localhost:9000',
```
vied12: Please, don't :sweat_smile:

server/settings.py
```diff
@@ -194,7 +194,7 @@ def env(variable, fallback_value=None):
 else:
     INSTALLED_APPS.append('apps.auth.db')
 
-SUPERDESK_TESTING = (env('SUPERDESK_TESTING', 'false').lower() == 'true')
+SUPERDESK_TESTING = (env('SUPERDESK_TESTING', 'true').lower() == 'true')
```
vied12: Can I know why please?

client/protractor-conf.js
vladnicoara: wow, committed the wrong file, just use this locally

server/settings.py
vladnicoara: same here, this file went to the commit by mistake

vied12:
When there is 0 blog, it shows `0 blogs`. It should be `0 blog`

nistormihai:
@vied12 you see why gettext is needed for pluralization
PS: gettext is supporting up to 5 messages for plural form, in `fr` I thing there are 3 forms, and it is not a problem if `0 blogs` in `fr` can be `0 le blog`

client/spec/themes_manager_spec.js
```diff
@@ -12,8 +12,6 @@ describe('Themes Manager', function() {
         .themes
         .then(function(themes) {
             expect(themes.length).toBe(3);
-            // angular-base
-            themeManager.expectTheme(0, {number_of_blogs_expected: 0, is_default_theme:false, name: 'Angular Base'});
```
vied12: We need to keep this check

PR #206, opened 2015-09-15
LBSD-649 added published_date for submitted posts on publish action

ancafarcas:


server/liveblog/posts/posts.py
```diff
@@ -143,7 +143,8 @@ def on_update(self, updates, original):
         post.update(updates)
         self.check_post_permission(post)
         # put the published item from drafts at the top of the timeline
-        if updates.get('post_status') == 'open' and original.get('post_status') == 'draft':
+        if (updates.get('post_status') == 'open' and original.get('post_status') == 'draft'
```
vied12: The parentheses seem wrong to prevent an unwanted operation priority here.
But anyway, I would simplify it like that:
```python
if updates.get('post_status') == 'open' and original.get('post_status') in ('draft', 'submitted'):
```

ancafarcas: ok, I will follow your sugestion but I verified several times and there was no priority problem there

vied12:
LGTM :checkered_flag:

PR #207, opened 2015-09-16
Feature/publish a contribution LBSD-632 LBSD-661

vied12:
New:

- A publisher can publish a *contribution*
- When a publisher unpublish a post it goes to the *contributions*
- User without the right to publish doesn't see the buttons to change a published post in the *timeline*.
- **[General]** Posts list permissions were refactored to allow function which receive a post as parameter.
example for contribution posts list declaration. See the **lb-posts-allow-*** bellow
```html
<lb-posts-list
  lb-posts-blog-id="blog._id"
  lb-posts-status="submitted"
  lb-posts-order-by="updated_first"
  lb-posts-allow-unpublishing="false"
  lb-posts-allow-reordering="false"
  lb-posts-allow-editing="blogSecurityService.isUserOwnerOrAdmin"
  lb-posts-allow-deleting="blogSecurityService.isUserOwnerOrAdmin"
  lb-posts-allow-publishing="blogSecurityService.canPublishAPost">
```


ancafarcas:
LGTM

client/app/scripts/liveblog-edit/directives.js
```diff
@@ -29,8 +29,11 @@ define([
                     angular.extend(vm, {
                         isLoading: true,
                         blogId: $scope.lbPostsBlogId,
-                        allowUnpublish: $scope.lbPostsAllowUnpublish,
-                        allowReordering: $scope.lbPostsAllowReordering,
+                        allowUnpublishing: $scope.lbPostsAllowUnpublishing || false,
```
nistormihai: what is the problem if it is `undefined`?

client/app/scripts/liveblog-edit/module.js
```diff
@@ -480,18 +480,17 @@ define([
...
+        }
+        function isUserOwnerOrAdmin(archive) {
+            return $rootScope.currentUser._id === archive.original_creator || isAdmin();
         }
         function goToSettings() {
             var def = $q.defer();
             blogService.get($route.current.params._id)
             .then(function(response) {
-                if (isUserOwner(response)) {
+                if (isUserOwnerOrAdmin(response)) {
```
nistormihai: IMHO just `hasRights` should be much better, because chaining all the right in the method name implies that in the future if something is changed the name will change also.

client/app/scripts/liveblog-edit/views/main.html
```diff
@@ -127,8 +132,10 @@ <h3 class="page-nav-title">{{ blog.title }}</h3>
                       lb-posts-list
                       lb-posts-blog-id="blog._id"
                       lb-posts-status="open"
-                      lb-posts-allow-unpublish="true"
-                      lb-posts-allow-reordering="true"
+                      lb-posts-allow-unpublishing="blogSecurityService.canPublishAPost"
```
nistormihai: all this mapping looks ugly,
pass the `blogSecurityService` to the directive or
a convention to make all the methods name starting with `can` to merge in the directive

server/liveblog/posts/posts.py
```diff
@@ -158,9 +162,10 @@ def on_updated(self, updates, original):
         # send notifications
         if updates.get('deleted', False):
             push_notification('posts', deleted=True, post_id=original.get('_id'))
-        elif updates.get('post_status') == 'draft':
-            push_notification('posts', drafted=True, post_id=original.get('_id'),
-                              author_id=original.get('original_creator'))
+        # NOTE: Seems unsused, to be removed later if no bug appears.
```
nistormihai: we need to somehow target the update to the related `collections` ex:
don't update the contributions if the draft got updated.

nistormihai:
in rest ALGTM

client/app/scripts/liveblog-edit/directives.js
vied12: I want boolean for all privileges. If undefined, I assume that the privilege is not given.
I know that undefined can be used as a false value but I prefer an explicit false. Then here, a code reader can understand that we set default values.

nistormihai: it sound reasonable if you would check against boolean type, if not they don't have any meaning and it just makes the code harder to read.

server/liveblog/posts/posts.py
vied12: But It's not used currently. And this is not the purpose of this PR.

client/app/scripts/liveblog-edit/views/main.html
vied12: We need this mapping here... This is the aim of Angular to do the mapping in the DOM...
and all the method start with *allow-*...

client/app/scripts/liveblog-edit/module.js
vied12: We have different rights, we need to name them. If we change the roles later, it makes sense to update the role management.

client/app/scripts/liveblog-edit/directives.js
vied12: If it's more easy to read for you, I will remove the default values.

client/app/scripts/liveblog-edit/module.js
nistormihai: no it doesn't need to be reflected in every line, it is not DRY, where it could easily be.

server/liveblog/posts/posts.py
nistormihai: but this PR change the way that notification are send, so how it is not the purpose of it?

client/app/scripts/liveblog-edit/module.js
vied12: I don't understand what do you suggest.

server/liveblog/posts/posts.py
vied12: I commented it to let us able to use it later, when we will need it. For now this is useless, and incomplete.

client/app/scripts/liveblog-edit/module.js
nistormihai: to rename the method to `hasRights` see above, so it doesn't matter what checks are made in that method regards to method name.

client/app/scripts/liveblog-edit/views/main.html
nistormihai: ok, but still looks ugly.

vied12: https://github.com/blog/1943-how-to-write-the-perfect-pull-request

client/app/scripts/liveblog-edit/module.js
vied12: As you can see here https://github.com/superdesk/liveblog/pull/167
An owner or an admin can do different things. How do you plan to identify these rights ?
The technical specifications describe clearly all the roles. What can change in the future ?
https://twitter.com/dhh/status/623598101127897088

vied12: By the way, the purpose of this PR is not to re-factor the privileges service, If you have an idea in mind, please describe it in a document then we can discuss it more globally if you don't mind.

nistormihai: no I didn't have in mind to refactor the privileges, but that method name has changed and keeping in mind that, it means that may change again, it was simple in the beginning.

server/liveblog/posts/posts.py
nistormihai: ok, seems right let's go like that but let's keep in mind this issue.

client/app/scripts/liveblog-edit/module.js
vied12: I added 'orAdmin' to reflect more precisely what this function test. The name can't change any more since the name match now perfectly with what the function does.

client/app/scripts/liveblog-edit/views/main.html
vied12: I still don't understand what are the advantages.
Either we do the mapping into a new directive to link privileges to different types of post list, or we do it in the directive parameters, as I did.
I really prefer the second option, since this is more explicit which actions are allowed where you use the directive. Also it make the directive usable for other purposes without having to modify a service.

PR #211, opened 2015-09-16
fix: disable publish button if empty.

nistormihai:
LBSD-629

### DIFF
--- a/client/app/scripts/liveblog-edit/module.js
+++ b/client/app/scripts/liveblog-edit/module.js
@@ -56,6 +56,7 @@ define([
 
         // define the $scope
         angular.extend($scope, {
+            publishDisabled: true,
             blog: blog,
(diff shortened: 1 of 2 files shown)


client/app/scripts/liveblog-edit/module.js
```diff
@@ -56,6 +56,7 @@ define([
 
         // define the $scope
         angular.extend($scope, {
+            publishDisabled: true,
```
vied12: What is it for ? I don't see this variable used somewhere. Maybe did you forgot to use it in the view ?

nistormihai: True it is missing.
It is used to disable or enable the publish button, and it was there in the view.
It seems that a merge has modify it, I will add it.

PR #212, opened 2015-09-17
Switch project license over to AGPLv3

hlmnrmr:


vied12:
nice licence ! :ok_hand:

PR #219, opened 2015-09-22
fix: disable publish button if empty.

nistormihai:
LBSD-629


client/app/scripts/liveblog-edit/module.js
```diff
@@ -48,14 +48,17 @@ define([
         }
 
         // remove and clean every items from the editor
-        function cleanEditor() {
+        function cleanEditor(val) {
```
vied12: It is not clear for me what does represent the parameter `val` here.
Can't we find a better name ?

PR #220, opened 2015-09-22
feature: added link to the live blog output

nistormihai:
LBSD-625

### DIFF
--- a/client/app/scripts/liveblog-edit/module.js
+++ b/client/app/scripts/liveblog-edit/module.js
@@ -229,9 +229,9 @@ define([
     }
 
     BlogSettingsController.$inject = ['$scope', 'blog', 'api', 'blogService', '$location', 'notify',
-        'gettext', 'config', 'modal', '$q', 'upload'];
+        'gettext', 'modal', '$q', 'upload'];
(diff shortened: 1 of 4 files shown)


client/app/scripts/liveblog-edit/module.js
```diff
@@ -246,9 +247,7 @@ define([
                 }
                 vm.tab = tab;
             },
-            // take the public url (from s3) or the local address
-            // FIXME: The local address shouldn't be given on production mode
-            iframe_url: blog.public_url || config.server.url.replace('/api', '/embed/' + blog._id),
```
vied12: If you remove this line, I think that `config` isn't needed any more here, you can remove it from the dependencies

client/app/scripts/liveblog-edit/views/main.html
```diff
@@ -135,6 +140,6 @@ <h3 class="page-nav-title">{{ blog.title }}</h3>
                     </div>
                 </div>
             </div>
-        </div>
+        </div>
```
vied12: :trollface:

vied12:
LGTM :)

PR #223, opened 2015-09-24
Fix/close order mode

nistormihai:


client/app/scripts/liveblog-edit/directives.js
```diff
@@ -205,18 +212,29 @@ define([
                                 });
                             },
                             preMovePost: function(post) {
+                                $document.bind('keypress', escClearReorder);
```
vied12: Cool ! We need more shortcuts :)

client/app/scripts/liveblog-edit/directives.js
```diff
@@ -181,6 +187,7 @@ define([
                         }
 
                         angular.extend(scope, {
+                            // if the escape key is press then clear the reorder action.
```
vied12: Seems this comment is lost :)

client/app/scripts/liveblog-edit/views/posts.html
```diff
@@ -9,10 +9,11 @@
       index=$index
       posts-list-ctrl="postsList"
       start-reorder="postsList.startReorder(post)"
+      clear-reorder-action="postsList.clearReorder"
```
vied12: Later, it would be nice to remove these parameters and call directly method through `postsListCtrl` defined just above.
Because these parameters are not really part of the directive configuration.

vied12:
LGTM

PR #225, opened 2015-09-24
Feature/submitted notifications

vladnicoara:


client/app/scripts/liveblog-edit/module.js
```diff
@@ -63,6 +62,9 @@ define([
             selectedUsersFilter: [],
             currentPost: undefined,
             blogSecurityService: blogSecurityService,
+            postsService: {
+                getUnreadContributtions: postsService.getUnreadContributtions
```
vied12: typo: contributions takes one `t`

client/app/scripts/liveblog-edit/module.js
```diff
@@ -135,6 +137,11 @@ define([
                 $scope.panelState = panel;
                 // update url for deeplinking
                 $route.updateParams({panel: $scope.panelState});
+                //clear the new contribution notification
+                if (panel === 'contributions') {
+                    postsService.resetUnreadContributtions(panel);
+                }
+                postsService.setPanelState(panel);
```
vied12: This instruction is redundant with `$scope.panelState = panel;` L137,
and more generally, the posts service shouldn't be aware of the panels.
Panels are graphical elements, that only controllers should be aware of.

ancafarcas:
it LGTM

client/app/scripts/liveblog-edit/module.js
vladnicoara: I don't understand which instruction is redundant with '$scope.panelState = panel; L137,'
The service should be aware of the panels because the number of unread contributions should only be increased if a panel other than the contributions one is active

vied12: I agree with that, the number of unread posts depends of  course of what panel you opened. But it's not a good reason to let the post service to be aware of the panels.
Maybe you can put your method and post event listener directly into the controller, or create a new service for unread posts if you think that we will need it from more than one controller.

client/app/scripts/liveblog-edit/module.js
```diff
@@ -63,6 +63,9 @@ define([
             selectedUsersFilter: [],
             currentPost: undefined,
             blogSecurityService: blogSecurityService,
+            unreadPostsService: {
```
vied12: `unreadPostsService: unreadPostsService` may be enough

vladnicoara: it is enough but there is no need to overexpose

vied12:
LGTM :sweat_smile:

PR #229, opened 2015-09-28
Feature/update publish/unpublish icons LBSD-674

vied12:


nistormihai:
LGTM

nistormihai:
It would be nice that in the future we use vectors instead of sprites

PR #231, opened 2015-09-28
feature: disable reorder for single post on a blog.

nistormihai:
LBSD-612

### DIFF
--- a/client/app/scripts/liveblog-edit/directives.js
+++ b/client/app/scripts/liveblog-edit/directives.js
@@ -98,6 +98,9 @@ define([
                         isPostsEmpty: function() {
                             return vm.pagesManager.count() < 1 && !vm.isLoading;
                         },
+                        isSinglePost: function() {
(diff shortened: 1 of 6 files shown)


client/app/scripts/liveblog-edit/directives.js
```diff
@@ -98,6 +98,9 @@ define([
                         isPostsEmpty: function() {
                             return vm.pagesManager.count() < 1 && !vm.isLoading;
                         },
+                        isSinglePost: function() {
+                            return vm.pagesManager.count() === 1 && !vm.isLoading;
```
vied12: Aren't we mixing different behaviour here ? 
Is it the role of the function `isSinglePost` to return false if the controller is, for any reason, loading, even if there is actually one single post ?

vladnicoara:
LGTM

vied12:
IDEM :ok_hand:

PR #233, opened 2015-09-30
Feature/request liveblog membership

ancafarcas:


vied12:
Why this new feature is dispatched into 3 functions `notify_the_owner`, `send_email_to_owner` and `send_owner_email`? Do we need to call one of them somewhere else ?
The names are very similar and confusing. I would merge this feature into one function `request_membership`.
I would also remove this `origin` parameter. From what I see, there is no need to change it. You can retrieve the value of `app.config['CLIENT_URL']` within the function.

We also need to find a way to trigger this feature. By adding a new endpoint `blogs/<id>/request-membership` ?

ancafarcas:
The three similar functions are implemented after a previous implemented "send mail to members" model. I was thinking to be consequent with what we have.
One of these functions: "notify_the_owner" is triggered in find_one method(for now, only to test if it is working) but this is the thing I wanted to talk to you. We have to find a way to trigger it from elsewhere. I would agree with "blog/_id/request_membership" endpoint.

server/liveblog/blogs/blogs.py
```diff
@@ -144,6 +147,7 @@ def on_created(self, docs):
 
     def find_one(self, req, **lookup):
         doc = super().find_one(req, **lookup)
+#         notify_the_owner(doc, app.config['CLIENT_URL'])
```
vied12: if you have a  good reason to keep commented, it's better to add a note to let us know why it is important

server/liveblog/blogs/blogs.py
```diff
@@ -100,8 +103,8 @@ def send_members_email(recipients, user_name, doc, title, url):
     admins = app.config['ADMINS']
     app_name = app.config['APPLICATION_NAME']
     subject = render_template("invited_members_subject.txt", app_name=app_name)
-    text_body = render_template("invited_members.txt", link=url, title=title)
-    html_body = render_template("invited_members.html", link=url, title=title)
+    text_body = render_template("invited_members.txt", app_name=app_name, link=url, title=title)
+    html_body = render_template("invited_members.html", app_name=app_name, link=url, title=title)
```
vied12: This mail to invite a member shouldn't be part of the new `blog_membership` module ?

server/liveblog/blogs/blog_membership.py
```diff
@@ -0,0 +1,101 @@
...
+from flask import current_app as app, render_template
+from superdesk.emails import send_email
+from liveblog.blogs.blogs import BlogService
+from bson.objectid import ObjectId
+from superdesk.notification import push_notification
+
+logger = logging.getLogger('superdesk')
+
+
+def notify_the_owner(doc, origin):
```
vied12: The same remark as before about the names of these functions `notify_the_owner`, `send_email_to_owner` and `send_owner_email`. The names are confusing, and I guess that we can merge `send_email_to_owner` and `send_owner_email`

server/liveblog/blogs/blogs.py
vied12: would it be great to put all the features relative to membership in the same module ?
Otherwise we should rename this one by "request_membership" to avoid confusion

server/liveblog/blogs/blogs.py
ancafarcas: This comment it is supposed to be removed

server/liveblog/blogs/blogs.py
ancafarcas: I would go with renaming the newest module "request_membership" rather than moving the "invite_members" feature.  This invitation is sent on blog creation, if a member is added at that time, so we should keep it together with "blogs" features

vied12:
Otherwise, looks good ! :+1:

server/liveblog/blogs/blog_membership.py
ancafarcas: I agree with the merge of the two, it doesn't look as bad as I expect, if they are together

PR #236, opened 2015-09-30
iframe-resizer added to embed LBSD-684

vied12:
to allow dynamic height for iframe depending of the content


server/liveblog/client_modules/client_modules.py
```diff
@@ -36,7 +37,7 @@ class ClientBlogsResource(BlogsResource):
     schema.update(BlogsResource.schema)
 
 
-class ClientBlogsService(BlogService):
+class ClientBlogsService(BaseService):
```
nistormihai: what all above have to do with iframe-resizer?

server/liveblog/themes/themes_assets/angular-base/theme.json
```diff
@@ -11,6 +11,7 @@
         "//code.angularjs.org/1.4.2/angular-resource.js",
         "//cdnjs.cloudflare.com/ajax/libs/moment.js/2.10.3/moment.min.js",
         "vendors/lodash.min.js",
+        "vendors/iframeResizer.contentWindow.min.js",
```
nistormihai: I think that we should keep angular-base light as possible.

server/liveblog/themes/themes_assets/default-theme/main.js
```diff
@@ -59,6 +59,14 @@
         .then(function() {
             vm.permalinkScroll();
             $interval(retrieveUpdate, UPDATE_EVERY);
+            // listen events from parent
```
nistormihai: can we have a different theme with this iframe functionality?

server/liveblog/client_modules/client_modules.py
vied12: It resolves a bug that broke the embed route, because of a new permission check introduced on `devel`. 
Without this, I can't test the embed when I'm based on `devel`.

nistormihai:
In rest everything looks good and a very nice ideea :100: 
PS: we really need to add bower on dependencies, the list became big.

server/liveblog/themes/themes_assets/default-theme/main.js
vied12: Yes, I guess.
Do you mean that we should put this listener in `angular-base` ?

server/liveblog/client_modules/client_modules.py
jey07ro: please make another commit,
when searching in commit history this will not be clear why it is here.

nistormihai: please make another commit,
when searching in commit history this will not be clear why it is here.

PR #237, opened 2015-09-30
Fix/responsive modal

darconny:


client/app/scripts/liveblog-bloglist/styles/liveblog-bloglist.less
```diff
@@ -138,54 +138,54 @@
...
-    }
-    .member-list ul > li i {
-        position: absolute;
-        right: 10px;
-        top: 10px;
-        display: none;
-        opacity: .6;
-        filter: alpha(opacity=60);
-    }
+	.search-result {
```
vied12: On this project we indent with 4 spaces everywhere (less, js, python...).
Here you indented with tabulation

vied12:
It looks that the modal for adding team member is now broken on http://liveblogfixresponsivemodal.dev.superdesk.org/#/liveblog/

selection_082

Linux, Firefox 41.0

darconny:
Latest changes are not yet applied to the instance, for me all seems fine in FF.

client/app/scripts/liveblog-bloglist/views/main.html
```diff
@@ -90,7 +89,8 @@ <h3 translate>Create New Blog</h3>
             <div class="upload-progress" ng-show="progress.width">
                 <div class="bar" style="width: {{ progress.width }}%;"></div>
             </div>
-            <div class="modal-body" style="height: 470px; max-height:470px;">
+            <div class="modal-body" style="min-height: 300px;">
```
vied12: For the next time you see one,
static style in html must not happen. :)
I know that you removed one of them, but, while we're at it, we should remove all of them.

darconny: Yes you are right, but if you get bored with adding classes.. :)

PR #238, opened 2015-10-01
Fix/reorder cancel button

vladnicoara:


client/app/scripts/liveblog-edit/styles/liveblog-edit.less
```diff
@@ -58,8 +58,8 @@
             }
         }
         .content--timeline {
-            top: 10px !important;
-            padding-top: @nav-height - 10px;
+            top: 48px !important;
+            //padding-top: @nav-height - 10px;
```
vied12: Comment should follow a note if you decide to keep it

vied12:
Cool !  it's a start to isolate the reorder feature from the postsList directive :)
LGTM

PR #244, opened 2015-10-14
Feature/use GitHub for themes [LBSD-644]

vied12:
Shows a link to the theme github repository and provide a link to download the archive


nistormihai:
We have git subtree in liveblog2 and it is a hack to maintain,
the issue with subtree is that you can also commit changes into liveblog repo not only in the themes repos.
this will get very complicated when handling even more repos for different themes, this from IMHO is not the proper solution.

nistormihai:
the code part looks good to me :star2:

PR #245, opened 2015-10-15
lbFluidIframe directive added to angluar-theme and used in classic-theme

vied12:
New directive LbFluidIframe added to lb-theme-angular which save the aspect ratio of the orignal iframe, set the full width and preserve the original ratio.


nistormihai:
LGTM :star2:

vladnicoara:
LGTM

PR #246, opened 2015-10-15
Feature/request blog access

vladnicoara:


client/app/scripts/liveblog-bloglist/module.js
```diff
@@ -12,6 +12,7 @@
         ];
         $scope.activeState = isArchivedFilterSelected ? $scope.states[1] : $scope.states[0];
         $scope.creationStep = 'Details';
+        $scope.requestAccessMessage = gettext('Click to request access');
```
vied12: If you don't need to do any action on this string with javascript, the html file would be probably a better place.

vied12:
Otherwise, :+1:

client/app/scripts/liveblog-bloglist/module.js
vladnicoara: I am using it as a tooltip for a <li> element and as part of an expression something like title="{{ !isUserAllowedToOpenBlog(blog) ? requestAccessMessage: '' }}" so it's kindof hard to translate directly in html

vied12: Ah ok!
because `{{ !isUserAllowedToOpenBlog(blog) ? 'Click to request access' | translate : '' }}` wouldn't work ?

vladnicoara: nope, not working... just tried it.

vied12: haha, thank you for having tried :)

PR #249, opened 2015-10-19
  LBSD-701 added email notification for blog members

vied12:
Still need to filter the email receivers to let only new members to receive membership email


vied12:
:+1:

PR #250, opened 2015-10-20
Fix/ipad fix

darconny:
- fixed and tested on iPad 4 retina, initial scale set to 1 and editor/timeline are 50% width each
- also added flexbox prefixes that fix the editor buttons


vied12:
:+1:

nistormihai:
:iphone: :star:

PR #251, opened 2015-10-20
LBSD-703 list of users that have requested access to a blog

ancafarcas:


server/liveblog/blogs/request_membership.py
```diff
@@ -89,8 +87,20 @@ def on_created(self, docs):
         for doc in docs:
             push_notification(self.notification_key, created=1, request_id=str(doc.get('_id')))
         # and members with emails
-        notify_the_owner(docs, app.config['CLIENT_URL'])
+#             recipients = doc.get('original_creator', [])
```
vied12: please remove commented code if you don't need it :bowtie:

vied12:
otherwise: :ok_hand:

server/liveblog/blogs/request_membership.py
ancafarcas: I checked for commented code three times!!! It will be removed!

PR #253, opened 2015-10-23
priveleges needed for posting in request_membership

ancafarcas:


server/liveblog/blogs/request_membership.py
```diff
@@ -72,17 +71,19 @@ class MembershipResource(Resource):
...
-class MembershipService(BlogService):
-    notification_key = 'request'
 
+class MembershipService(BaseService):
+    notification_key = 'request'
+  
     def on_create(self, docs):
         for doc in docs:
+            doc['original_creator'] = str(get_user().get('_id'))
             doc['message'] = "Please add me as a contributor to your blog"
```
vied12: Do we really need to define the message here ? where is it used ?

ancafarcas: It isn't really needed. I think it can be deleted if @vladnicoara says it doesn't need it

vladnicoara: if you are just using that to let front-end know what kind of request it is and we only get one kind of request from that endpoint, I can do without it

ancafarcas: ok, removed it is, then!

PR #254, opened 2015-10-27
request for membership can be sent only one time

ancafarcas:


server/liveblog/blogs/request_membership.py
```diff
@@ -80,6 +82,12 @@ class MembershipService(BaseService):
     def on_create(self, docs):
         for doc in docs:
             doc['original_creator'] = str(get_user().get('_id'))
+            request_service = get_resource_service('request_membership').get(req=None, lookup={})
```
vied12: never do your filter operations in python when the database can do it for you :)
Use the `lookup` field.

ancafarcas: I used it but it didn't worked corectly

PR #258, opened 2015-10-29
Feature/preview editor

ancafarcas:


vied12:
I got two issues on the test instance:

* [ ] The tooltip is covered by the upper bar

selection_085
* [ ] When you edit an existing post, it's impossible to preview it
selection_086

client/app/scripts/liveblog-edit/views/main.html
```diff
@@ -92,11 +92,21 @@ <h3 class="page-nav-title">{{ blog.title }}</h3>
...
                 <!-- editor -->
-                <div class="panel--editor" ng-show="panelState === 'editor'">
-                    <header>
-                        Editor<button class="btn btn-info pull-right" ng-click="togglePreview()" translate>{{ preview? 'Edit': 'Preview' }}</button>
-                        <button class="btn pull-right" ng-click="askAndResetEditor()" translate>Reset</button>
-                    </header>
+                    <div class="panel--editor" ng-show="panelState === 'editor'">
+                        <div ng-if="!preview">
+                        <header>
+                            Editor<button class="btn btn-info pull-right" ng-click="togglePreview()" ng-disabled="actionStatus() && !isCurrentPostPublished()" translate>Preview
```
vied12: why `ng-disabled="actionStatus() && !isCurrentPostPublished()"` regarding the preview button ?
Does it mean that we can't see the preview if the post hasn't been published ? Why ?

client/app/scripts/liveblog-edit/styles/liveblog-edit.less
```diff
@@ -580,6 +580,13 @@
 }
 
 // preview changes
+
+.close-preview {
+    font-size: 2em;
+    color: #8e8e8e!important;
```
vied12: I'm not a big fan of `!important` declarations. Mostly outside of a scope.
And you will see that if you put it in a scope, you won't need anymore to use `!important` ;)
Everything in a module style should belong to the module scope, like a common parent class. For global style it's better to put it in `client/app/styles/less/liveblog.less`

client/app/scripts/liveblog-edit/views/main.html
vladnicoara: This && !isCurrentPostPublished() only keeps the preview button enabled when we are in editing mode. Before this, when we were editing an existing post the button was not enabled before making any changes. Nothing else is effected

vied12: well then this function `isCurrentPostPublished()` isn't really doing what its name says.

vladnicoara: isCurrentPostPublished() does what it says. The && !isCurrentPostPublished() also requires the post to NOT be published in order for the preview button to be disabled. That means that if the post is already published, the preview button will always be enabled

vied12:
:+1:

PR #260, opened 2015-11-02
LBSD-709 Fix inconsistency in the permission of an editor

vladnicoara:
Editors should be able to edit/remove/publish posts by any contributor
regardless of their role (i.e. from contributors, other editors, and
admins).


vied12:
But contributor should still be able to edit his own contributions, isn't it ?
Then you should create a new permission `isUserOwnerOrEditor` instead of using `canPublishAPost`

vied12:
LGTM :+1:

PR #261, opened 2015-11-04
Feature/big images

nistormihai:


client/app/scripts/liveblog-users/module.js
```diff
@@ -0,0 +1,151 @@
...
+                    $scope.notify.message = (error.statusText !== '') ? error.statusText : gettext('There was a problem with your upload');
+                    $scope.notify.code = 'error';
+                }, function(update) {
+                    $scope.progress.width = Math.round(update.loaded / update.total * 100.0);
+                });
+            });
+        };
+    }
+
+    angular.module('superdesk.users')
```
vied12: Did you copy this module from superdesk ?
There is no other way to limit code duplication ? This is something that we want to avoid, when we update superdesk for instance :)

nistormihai: Well actually this controller had been modified from the one in superdesk

vied12: Then you should update the original module on the SD repository.  Otherwise
think about me when I will have to update it :)

nistormihai: It is a different implementation than the superdesk,
Update should be affected, unless the ChangeAvatarController controller is changed in a way that is not compatible with the old one.

vied12:
LGTM

PR #263, opened 2015-11-09
LBSD-718 archive endpoint fo blog image upload

ancafarcas:


client/app/scripts/liveblog-bloglist/module.js
```diff
@@ -75,7 +75,7 @@
             }
             // return a promise of upload which will call the success/error callback
             return api.upload.getUrl().then(function(url) {
```
vied12: you still use the upload url here

vied12:
:+1:

PR #266, opened 2015-11-10
custom generic embed for themes

vied12:
selection_085

### DIFF
--- a/server/liveblog/themes/themes_assets/angular/liveblog-embed/directives.js
+++ b/server/liveblog/themes/themes_assets/angular/liveblog-embed/directives.js
@@ -49,5 +49,31 @@
                     elem.html(attrs.lbTwitterContent);
                 }
             };
-        }]);
+        }])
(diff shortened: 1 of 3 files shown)


server/liveblog/themes/themes_assets/angular/liveblog-embed/directives.js
```diff
@@ -49,5 +49,31 @@
...
             };
-        }]);
+        }])
+        .directive('lbGenericEmbed', ['$timeout', function($timeout) {
+            return {
+                scope: {
+                    item: '='
+                },
+                template: [
+                    '<article class="item--embed" ng-click="item.meta.url">',
```
vladnicoara: Any reason you used inline templates?

vied12: this is faster to load for the client. But later we can setup a gulp/grunt task to compile all the template files with `$templateCache`.

vladnicoara:
LGTM

PR #267, opened 2015-11-12
embed-specific message when the themes are not registered

ancafarcas:


server/liveblog/embed/embed.py
```diff
@@ -104,6 +105,10 @@ def embed(blog_id, api_host=None, theme=None):
     # retrieve the wanted theme and add it to blog['theme'] if is not the registered one
     try:
         theme_name = request.args.get('theme', theme)
+        theme = get_resource_service('themes').find_one(req=None, name=blog['blog_preferences'].get('theme'))
```
vied12: be careful with duplication, this line already exists below.
You should check if it's empty on the one below to prevent another call to the database.

ancafarcas: I think we should remove the one below now that we checked theme existence above. If it's empty we already have introduced the error message.

vied12: You can't because here you are overwriting the `theme` variable given as parameter.
This parameter is useful to overwrite the theme set in the settings.

ancafarcas: o.k., I'll keep it and only do the checkup but I'm trying to say that when we did the checkup 
if theme is None:
            raise SuperdeskApiError.badRequestError(...)
we made sure that if we have "None" value for the theme, it won't pass further. On the other hand, if we do have a theme we continue to use the theme we obtained above using the line no. 108 . Not true?

vied12: But it should continue even if there is no theme defined in preferences because an other theme can be asked through the `theme` parameter.

server/liveblog/embed/embed.py
```diff
@@ -104,11 +105,15 @@ def embed(blog_id, api_host=None, theme=None):
...
     try:
         theme_name = request.args.get('theme', theme)
+
     except RuntimeError:
         # this method can be called outside from a request context
         theme_name = theme
-    # collect static assets to load them in the template
+
     theme = get_resource_service('themes').find_one(req=None, name=blog['blog_preferences'].get('theme'))
+    if theme is None:
```
vied12: you also need to check if `theme_name` is **not** `None` here, otherwise, the fall-back from parameter won't work

vied12:
:ok_hand:

PR #269, opened 2015-11-17
LBSD-738 back-end support for comments

ancafarcas:


server/settings.py
```diff
@@ -185,6 +185,9 @@ def env(variable, fallback_value=None):
 S3_USE_HTTPS = False
 FLASK_ASSETS_USE_S3 = False
 USE_S3 = FLASK_ASSETS_USE_S3
+RECAPTCHA_PUBLIC_KEY = ''
+RECAPTCHA_PRIVATE_KEY = ''
```
vied12: it's easier later to assign value if you use the `env()` function. No need to declare empty string as default

server/liveblog/client_modules/client_modules.py
```diff
@@ -59,6 +63,53 @@ class ClientPostsService(PostsService):
     pass
 
 
+class ClientItemsResource(ItemsResource):
```
vied12: Why do we need items here ?
Can we create multiple items comment ?
If no, can't we have just this `Comment` resource below ?

server/requirements.txt
```diff
@@ -5,3 +5,4 @@ git+git://github.com/joehand/flask-s3@8fe9f27ebb5251b7cf53a6684595d41dc4af2fbf#e
 # Superdesk
 -e git+git://github.com/liveblog/superdesk-core@a117d79e7fa10c92acf17aff957d05fdf2ac3f9b#egg=Superdesk-Core==0.0.1-dev
 -e git+git://github.com/liveblog/superdesk@33029d9fd55f7c7047a60abdac1f1911f044a27a#egg=Superdesk-Server
+recaptcha-client==1.0.6
```
vied12: This library doesn't seem compatible with python3

server/liveblog/posts/posts.py
```diff
@@ -87,6 +88,12 @@ class PostsResource(ArchiveResource):
             'type': 'datetime'
         },
         'publisher': Resource.rel('users', True),
+        'recaptcha_challenge_field': {
```
vied12: Do we really need to save it in the database ? Is it used later ?

server/liveblog/client_modules/client_modules.py
ancafarcas: We need Items resource because we follow the exact same structure as for posts. So, a post is a package which contains one ore multiple items. Here a comment is a package which contains an item. It doesn't make sense to post an empty package (that is why we need items)

server/liveblog/posts/posts.py
ancafarcas: No, this one is not needed and it will be removed

server/settings.py
```diff
@@ -185,6 +185,9 @@ def env(variable, fallback_value=None):
 S3_USE_HTTPS = False
 FLASK_ASSETS_USE_S3 = False
 USE_S3 = FLASK_ASSETS_USE_S3
+RECAPTCHA_PUBLIC_KEY = env('RECAPTCHA_PUBLIC_KEY', '')
```
vied12: I don't think we need these empty string as default value.
It's better to keep `None`

server/requirements.txt
```diff
@@ -5,3 +5,4 @@ git+git://github.com/joehand/flask-s3@8fe9f27ebb5251b7cf53a6684595d41dc4af2fbf#e
 # Superdesk
 -e git+git://github.com/liveblog/superdesk-core@a117d79e7fa10c92acf17aff957d05fdf2ac3f9b#egg=Superdesk-Core==0.0.1-dev
 -e git+git://github.com/liveblog/superdesk@33029d9fd55f7c7047a60abdac1f1911f044a27a#egg=Superdesk-Server
+norecaptcha-python3
```
vied12: should be removed

nistormihai: @ancafarcas is this pr ready to be merge into my branch?

ancafarcas: yes, it can be merged now

vied12:
https://github.com/superdesk/liveblog/pull/299 is already implementing comments.
Can this PR be closed ?

ancafarcas:
It can be closed from my part if @nistormihai confirm it.

vied12:
Duplicated #299

PR #273, opened 2015-11-19
Fix/add content here style

darconny:
LBSD-739
Editor - adjust style For "Add content here" button to be better visible


vied12:
:+1:

PR #275, opened 2015-11-19
Store the post order sequence in database instead of redis LBSD-708

vied12:
depends on https://github.com/superdesk/superdesk-core/pull/86


ancafarcas:
:+1:

PR #278, opened 2015-11-26
feature: implement image rendition for blog image in the embed.

nistormihai:
use srcset in embed for blog image with the settings rendition.

### DIFF
--- a/server/liveblog/themes/themes_assets/classic/main.js
+++ b/server/liveblog/themes/themes_assets/classic/main.js
@@ -16,6 +16,11 @@
         function retrieveUpdate() {
             return vm.pagesManager.retrieveUpdate(true);
         }
+        var srcset = '';
+        angular.forEach(config.blog.picture.renditions, function(value) {
+            srcset += ', ' + value.href + ' ' + value.width + 'w';
+        });
+        config.blog.picture_srcset = srcset.substring(2);
         // define view model
         angular.extend(vm, {
             templateDir: config.assets_root,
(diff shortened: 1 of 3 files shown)


server/liveblog/themes/themes_assets/classic/main.js
```diff
@@ -16,6 +16,11 @@
         function retrieveUpdate() {
             return vm.pagesManager.retrieveUpdate(true);
         }
+        var srcset = '';
+        angular.forEach(config.blog.picture.renditions, function(value) {
+            srcset += ', ' + value.href + ' ' + value.width + 'w';
```
vied12: Why not using the form 
```
srcset="image.jpg 1x,
             image-320.jpg 320w 1x,
             image-640.jpg 320w 2x"
```
since we have a `max-width` fixed, in this way it would support retina screen

jey07ro: https://ericportis.com/posts/2014/srcset-sizes/

nistormihai: https://ericportis.com/posts/2014/srcset-sizes/

vied12: ... TLTR
please explicit your thoughts...

jey07ro: the explicit thoughts are there, size and ratio are not a good idea because browser will not know which ones should it peek. also I don't know if that is valid syntax.

nistormihai: the explicit thoughts are there, size and ratio are not a good idea because browser will not know which ones should it peek. also I don't know if that is valid syntax.

vied12:
:+1:

PR #287, opened 2015-12-02
LBSD-737 update a theme then re- publish blogs

ancafarcas:


vied12:
The blogs that are using children themes should be republished too.
WIP

PR #289, opened 2015-12-02
LBSD-740 fixed the notification messages

vladnicoara:


vied12:
LGTM

In the future I would like to see like in Slack, a countdown in the disconnection message before the next try :relaxed:

nistormihai:
LGTM

also I would like to see a manual action instead of a countdown.

PR #290, opened 2015-12-02
load blog without cache

vied12:


client/app/scripts/liveblog-edit/module.js
```diff
@@ -530,7 +530,7 @@ define([
     BlogResolver.$inject = ['api', '$route', '$location', 'notify', 'gettext', 'blogService'];
     function BlogResolver(api, $route, $location, notify, gettext, blogService) {
 
-        return blogService.get($route.current.params._id)
+        return blogService.get($route.current.params._id, undefined, false)
```
nistormihai: from what I see blogService.get doesn't have additional parameters then _id
https://github.com/superdesk/liveblog/blob/devel/client/app/scripts/liveblog-edit/blog.service.js

PR #294, opened 2015-12-04
LBSD-750 Add space before "updated" in timeline

vladnicoara:


nistormihai:
so css wasn't an option :8ball: 
LGTM

PR #297, opened 2015-12-10
Feature/center tweet card

vied12:
- [x] Align tweet card in the middle
-  [x] fix a js error with srcset when there is no illustration


vladnicoara:
LGTM

PR #299, opened 2015-12-14
Feature/embed comments

nistormihai:


client/app/scripts/liveblog-edit/styles/post.less
```diff
@@ -1,6 +1,11 @@
 @import 'mixins.less';
 @import 'variables.less';
 
+.lb-mainitem-comment .main-avatar, .lb-item-comment .avatar{
+    background-image: url("../../images/comments-timeline.svg") !important;
```
vied12: Can't we avoid these `!important` ?
You can also play with selector deepness

server/liveblog/items/items.py
```diff
@@ -50,6 +50,17 @@ class ItemsResource(ArchiveResource):
         'item_type': {
             'type': 'string'
         },
+        'name': {
+            'type': 'string',
+            'minlength': 1,
+            'maxlength': 300
+        },
+        'recaptcha_challenge_field': {
```
vied12: I don't think we need these fields any more

server/requirements.txt
```diff
@@ -5,3 +5,4 @@ git+git://github.com/joehand/flask-s3@8fe9f27ebb5251b7cf53a6684595d41dc4af2fbf#e
 # Superdesk
 -e git+git://github.com/liveblog/superdesk-core@a117d79e7fa10c92acf17aff957d05fdf2ac3f9b#egg=Superdesk-Core==0.0.1-dev
 -e git+git://github.com/liveblog/superdesk@33029d9fd55f7c7047a60abdac1f1911f044a27a#egg=Superdesk-Server
+recaptcha-client==1.0.6
```
vied12: to be removed

server/settings.py
```diff
@@ -185,6 +185,9 @@ def env(variable, fallback_value=None):
 S3_USE_HTTPS = False
 FLASK_ASSETS_USE_S3 = False
 USE_S3 = FLASK_ASSETS_USE_S3
+RECAPTCHA_PUBLIC_KEY = ''
```
vied12: to be removed

server/liveblog/items/items.py
ancafarcas: some files are already updated regarding back-end part, in the pr #269

server/liveblog/themes/themes_assets/classic/main.js
```diff
@@ -32,6 +33,30 @@
             loading: true,
             finished: false,
             showAuthor: SHOW_AUTHOR,
+            comment: {
```
vied12: This timeline controller is already quite big, I would have created a special controller for Comments, with also a new directive to combine controller and view. What do you think?

vied12:
LGTM, more comments in code would be awesome :relaxed:

vied12:
And what about e2e tests ?

server/liveblog/themes/themes_assets/angular/liveblog-embed/resources.service.js
```diff
@@ -40,20 +40,44 @@
 
     Posts.$inject = ['$resource', 'config', 'users'];
     function Posts($resource, config, users) {
+        function _completeUser(obj) {
+            if (obj.name) {
+                obj.original_creator = {display_name: obj.name};
+            } else if(obj.original_creator !== "" && obj.original_creator !== 'None'){
```
vied12: Why `None`? this should be `null` or `underfined`, but not `None` right ?
And no double quote in `obj.original_creator !== ""` please

client/app/scripts/liveblog-edit/posts.service.js
```diff
@@ -93,10 +102,24 @@ define([
...
-            // complete Post With User Information
-            userList.getUser(post.original_creator).then(function(user) {
-                post.user = user;
+            post.comments = _.reduce(post.groups[1].refs, function(is, val) {
+                return is || _.isUndefined(val.item.name);
+            }, false);
+            // check if `fullDetails` flag is needed
+            // comments items set falg to true.
+            post.fullDetails = post.comments;
+            // special cases for comments.
```
vied12: I don't think this is the best solution. I don't see why we do need a special case for comment.
The simplest rule should be: we show the updated date when the updated date is older than creation date.
The solution would be to prevent an updated date to appear when you add a comment to a post

client/app/scripts/liveblog-edit/posts.service.js
```diff
@@ -93,10 +102,24 @@ define([
...
             });
-            // complete Post With User Information
-            userList.getUser(post.original_creator).then(function(user) {
-                post.user = user;
+            post.comments = _.reduce(post.groups[1].refs, function(is, val) {
+                return is || _.isUndefined(val.item.name);
+            }, false);
+            // check if `fullDetails` flag is needed
+            // comments items set falg to true.
+            post.fullDetails = post.comments;
```
vied12: this is not clear why it is for.
This look again like a special case for comment :)
What problem does it solves ?

client/app/scripts/liveblog-edit/styles/post.less
```diff
@@ -1,6 +1,20 @@
 @import 'mixins.less';
 @import 'variables.less';
 
+.avatar-comment {
```
vied12: can it be included in a scope ? `.lb-post` maybe ?

server/liveblog/blogs/blogs.py
```diff
@@ -147,16 +147,18 @@ def on_created(self, docs):
...
-            members = [str(m['user']) for m in doc.get('members', [])]
-            # add owner id to members
-            members.append(doc.get('original_creator'))
-            # check if current user belongs to members, and raise an exeption if not
-            if str(get_user().get('_id')) not in members:
-                roles = get_resource_service('roles').find_one(req=None, _id=get_user().get('role'))
-                if not roles:
-                    raise SuperdeskApiError.forbiddenError(message='you do not have permission to open this blog')
+        user = get_user()
+        if user:
```
vied12: Why did you add this check?
Why a no authenticated use should access comments ?

server/liveblog/items/items.py
```diff
@@ -50,6 +50,16 @@ class ItemsResource(ArchiveResource):
         'item_type': {
             'type': 'string'
         },
+        'name': {
+            'type': 'string',
+            'minlength': 1,
+            'maxlength': 30
+        },
+        'content': {
```
vied12: what is `content`? what the difference with `text` ?

client/app/scripts/liveblog-edit/posts.service.js
```diff
@@ -157,7 +180,8 @@ define([
                             blog: blog_id,
                             text: item.text,
                             meta: item.meta,
-                            item_type: item.item_type
+                            item_type: item.item_type,
+                            name: item.meta && item.meta.name
```
vied12: `name` can be equals to `false` here. Is it possible in the model point of view ? It seems unnatural :)

server/liveblog/themes/themes_assets/classic/template.html
```diff
@@ -72,8 +83,42 @@ <h1 ng-if="timeline.settings.showTitle" ng-bind="timeline.blog.title"></h1>
...
+                            <div ng-if="!timeline.settings.authorNameLinksToEmail || post.mainItem.item_type === 'comment'"
+                                 class="lb-author__name"
+                                 ng-bind="post.mainItem.original_creator[timeline.settings.authorNameFormat]"></div>
+                            <div ng-if="timeline.settings.authorNameLinksToEmail && post.mainItem.item_type !== 'comment'"
+                                 class="lb-author__name">
+                                <a href="mailto:{{post.mainItem.original_creator.email}}"
+                                   ng-bind="post.mainItem.original_creator[timeline.settings.authorNameFormat]"></a>
+                            </div>
+                        </div>
+                        <div class="lb-item" ng-switch="item.item_type">
```
vied12: There is too much code duplication here. I think that we can factorize it with `<script/>' templates or directives.

server/liveblog/themes/themes_assets/classic/template.html
```diff
@@ -85,7 +130,8 @@ <h1 ng-if="timeline.settings.showTitle" ng-bind="timeline.blog.title"></h1>
                     <article ng-switch-default lb-bind-html html-content="{{ item.text }}"></article>
                 </div>
             </div>
-            <div class="lb-updated-date ng-cloak" ng-show="post._updated !== post.published_date">
+
+            <div class="lb-updated-date" ng-show="post.showUpdate">
```
vied12: did you removed `ng-cloak` on purpose ?

nistormihai: merge issues

client/app/scripts/liveblog-edit/posts.service.js
nistormihai: `fullDetails` it is used to check if the post needs information on item.
`comments` it is used to see if the post has comments on it.

`comments` flag it is also used to compile if it is needed the *update time*

client/app/scripts/liveblog-edit/posts.service.js
nistormihai: it is specified in the task.

server/liveblog/themes/themes_assets/angular/liveblog-embed/resources.service.js
nistormihai: it comes from backend in string `None`.

client/app/scripts/liveblog-edit/styles/post.less
nistormihai: yes.

server/liveblog/blogs/blogs.py
nistormihai: @ancafarcas can you check this?

client/app/scripts/liveblog-edit/posts.service.js
vied12: Which task are you talking about ? Removing the date for post that contains a comment?
I'm talking about this one too, but I am suggesting another way to solve it, without making a special case here for comments. We agree on the problem I guess. It's about the solution here :)

server/liveblog/themes/themes_assets/angular/liveblog-embed/resources.service.js
vied12: but this is not fine, isn't it ?

client/app/scripts/liveblog-edit/posts.service.js
vied12: ok then I don't think it belongs to the post service if it's related to how we want to show it in a view.
Why not putting this logic into the view and showing comments details there ? Just like we can have different template for embedded code ?

client/app/scripts/liveblog-edit/posts.service.js
nistormihai: * if we publish a comment as a single-item post (comment-only) we
   * show the time of publishing (not comment creation)
   * and do not show the updated time when it is first published
* if we publish a comment in a multi-item post (comment + reaction from editors)
   * show time for each item
   * and not the updated time

vied12: I agree with the finish. I'm just talking about the code. I find it strange to have a special case for comments in the posts service because of our view needs. This logic should belongs to the view from my point of view.

client/app/scripts/liveblog-edit/posts.service.js
nistormihai: Business logic should belongs to the model from my point of view.

server/liveblog/themes/themes_assets/classic/template.html
nistormihai: I will take a look how this will be, and I think it makes sense.

client/app/scripts/liveblog-edit/posts.service.js
```diff
@@ -93,10 +102,24 @@ define([
                 mainItem: post.groups[1].refs[0],
                 items: post.groups[1].refs
             });
-            // complete Post With User Information
-            userList.getUser(post.original_creator).then(function(user) {
-                post.user = user;
+            post.comments = _.reduce(post.groups[1].refs, function(is, val) {
+                return is || _.isUndefined(val.item.name);
```
vied12: post gets comments if `val.item.name` is undefined for one of the item? I don't follow the logic :)
What is "name"? Can this expression be more clear ?

client/app/scripts/liveblog-edit/posts.service.js
vied12: Ok if this is business logic. Then you probably need to document these fields better and the names should be more explicit (`comments` here suggest a list but this is actually a boolean and the purpose of `fullDetails` is not obvious here).
Also there is a `angular.extends` block just above. I think you should move these new field attributions into it.

server/liveblog/items/items.py
ancafarcas: the `content` has the same role as text but here we have `maxlength` and `minlength` constraints

server/liveblog/blogs/blogs.py
ancafarcas: If there is a not authenticated user, the check-ups for `admin` and `members` of the blog will not take place and they will be able to access the comments since in the comments part we used client_blogs

server/liveblog/items/items.py
vied12: So it's not good for consistency :) We can't have two different fields for the same purpose in _Items_. You should do your constraints verification at the comment level. Or to create a sub model for comments based on _Item_

server/liveblog/blogs/blogs.py
vied12: sorry, I don't understand. Does an unauthenticated user can find a comment ? Because he shouldn’t. So the remaining question is still: why did you do that ? :)

ancafarcas: When a user is not authenticated then the `if is admin/member` check-up or if useless so I put that additional `if user` in order to give the users without credentials, possibility to post a comment

vied12: I understand that unauthenticated users must be able to post a comment, but here we are in `find_one` method. Why a user without credentials should be allowed to find a comment ?

server/liveblog/items/items.py
ancafarcas: I moved those two fields (name and content) inside `client_modules` module so that they are comments specific now

client/app/scripts/liveblog-edit/posts.service.js
nistormihai: `fullDetails` is obvious is full details on a post.
for `comments` I will change to `hasComments`.
`angular.extend` will not work here, and is a matter of preferences.

client/app/scripts/liveblog-edit/posts.service.js
nistormihai: well actually is `undefined` which is how it supposed to be.

client/app/scripts/liveblog-edit/posts.service.js
vied12: At this logic level, `fullDetails` is not meaning full. Does it means that all the details are filled ? Are they required ? Must they be shown ? And what are details?
Regarding `angular.extend`, it's a matter of preference, but here it has been already used above, so it would be more consistent to continue like that. It also prevent to assign a value twice (jshint will ring the bell)

client/app/scripts/liveblog-edit/posts.service.js
vied12: :ok_hand:

server/liveblog/blogs/blogs.py
ancafarcas: The `find_one` method is for finding a blog not a comment.

client/app/scripts/liveblog-edit/posts.service.js
nistormihai: let's slack this.

server/liveblog/blogs/blogs.py
vied12: right, this is for blog. But still, when I read this block, it looks like a security breach :)
I still don't understand why an unauthenticated user could find a blog when an authenticated user without permission can't do it

ancafarcas: I guess I couldn't find another way at that moment but now...if I think better, instead of `'blog': Resource.rel('blogs', True)` I should use `'blog': Resource.rel('client_blogs', True)` and it won't be needing a security hack

PR #313, opened 2016-01-19
themes settings at blog level

ancafarcas:
# FIXME

- [x] When I upload a theme with new setting value, it must be assigned to the saved theme if it still has the original value


server/liveblog/themes/themes.py
```diff
@@ -174,6 +174,28 @@ def save_or_update_theme(self, theme, files=[], force_update=False):
...
+            default_theme_settings = get_resource_service('themes').get_default_settings(theme)
+            # retrieve the default settings of the previous theme
+            default_prev_theme_settings = get_resource_service('themes').get_default_settings(previous_theme)
+
+            for blog in blogs:
+                for key, value in blog['theme_settings'].items():
+                    # if the values of theme setting from blog level are the same as the settings
+                    # from previous theme update the settings
+                    if value == default_prev_theme_settings[key]:
+                        blogs_service.system_update(ObjectId(blog['_id']),
```
vied12: It's better to not save the blog after each settings but at the end, before doing the same for the next blog.

server/liveblog/themes/themes.py
```diff
@@ -174,6 +174,28 @@ def save_or_update_theme(self, theme, files=[], force_update=False):
...
+                                                    {'theme_settings': default_theme_settings}, blog)
```
vied12: we don't want to keep all the new values, only the one for the key `key`

server/liveblog/themes/themes.py
```diff
@@ -174,6 +174,28 @@ def save_or_update_theme(self, theme, files=[], force_update=False):
...
+                    else:
+                        # keep the customized settings made on blog level and update the rest
+                        # with the settings of current theme
+                        default_theme_settings[key] = value
```
vied12: here, it's better to do nothing, since we want to keep the custom value

vied12:
since this feature is complex, it would be useful to have a behave test here, which fails without the feature

server/liveblog/themes/themes.py
ancafarcas: I agree

server/liveblog/themes/themes.py
ancafarcas: ok

vied12:
:+1:

server/features/client_modules.feature
```diff
@@ -6,9 +6,13 @@ Feature: Client modules operations
...
 
     Scenario: List blogs without needing auth
+    	Given "themes"
+        """
+        [{"name": "forest"}]
+        """
         Given "blogs"
         """
-        [{"title": "testBlog one"}, {"title": "testBlog two"}]
+        [{"title": "testBlog one", "blog_preferences": {"theme": "forest", "language": "fr"}}, {"blog_preferences": {"theme": "forest", "language": "fr"}, "title": "testBlog two"}]
```
vied12: `blog_preferences` is duplicated here

server/liveblog/blogs/blogs.py
```diff
@@ -133,6 +136,13 @@ def on_create(self, docs):
             prefs = global_prefs.copy()
             prefs.update(doc.get('blog_preferences', {}))
             doc['blog_preferences'] = prefs
+            doc['theme_settings'] = {}
```
vied12: is this line useful ?

ancafarcas: It initializes the empty dict where the themes settings will be saved on blog level. Yeah, I think we could live without it though :)

server/features/client_modules.feature
ancafarcas: actually it isn't but we have two blogs, each of them having their own `blog_preferences`

vied12: Yes of course, sorry !

ancafarcas:
TO DO: verify if the settings that were not customized (at blog level or at theme level), are being overwritten.

PR #316, opened 2016-01-28
timestamp created for drafts and contribution

ancafarcas:


nistormihai:
IT IS OK

vied12:
:+1:
Does the syntax `{{ (post._published || post._created) | reldate }}` not work ?
Otherwise I would create a method in the controller that returns the date instead of duplicating the tag.
As you want :)

ancafarcas:
The mentioned syntax does not work (it keeps the drafts/contribution created time even when the draft/contribution is published)

PR #317, opened 2016-02-04
LBSD-826 Hyperlinks don't always work from email clients( Thunderbird and such)

vladnicoara:


vied12:
We need this absolute url in other contexts than email. Would be better to do it at the email level instead.

nistormihai:
LGTM

PR #318, opened 2016-02-05
LBSD-827 Fix the about year

vladnicoara:
The permanent Javascript version that gets the current year breaks the modal window( and that may not even show the real date as it is easily manipulated by the client) so there... hardcoded


vied12:
:+1:

PR #319, opened 2016-02-05
sticky post status and highlight field

ancafarcas:


server/liveblog/posts/posts.py
```diff
@@ -170,7 +174,7 @@ def on_update(self, updates, original):
         post.update(updates)
         self.check_post_permission(post)
         # when publishing, put the published item from drafts and contributions at the top of the timeline
-        if updates.get('post_status') == 'open' and original.get('post_status') in ('draft', 'submitted'):
+        if updates.get('post_status') == 'open' and original.get('post_status') in ('draft', 'submitted', 'sticky'):
```
vied12: Do you think that `... original.get('post_status') not 'open':` could work and be more clear here ?

ancafarcas: it works with `original.get('post_status') not in ('open')` , but I realized that when we unpin the sticky post, it will go at the top of the timeline which is not expected since it should go in the timeline according to its timestamp. So, I made a correction for this purpose

vied12:
no need to use a tuple `('open')` if you have only one element.
Because how do you unpin a sticky post ?
if it's by setting `post_status` to `open`, it should change anything.
Why do you think it behaves differently ?

ancafarcas:
Yes, I thought that the tuple shouldn't be used by I obtained an error there if I didn't use it
For the unpining, the idea was to change the `post_status` from `sticky` to `open`

vied12:
:+1:

PR #323, opened 2016-02-10
fix: sir trevor add target blank.

nistormihai:


vied12:
I'm not sure this feature is a good idea since we have a filter that fix links (https://github.com/superdesk/liveblog/pull/322)
It adds complexity and disallow the theme developer to use another target (like `_parent`).
I would keep the link formatter minimalist as it is now and use only the filter to convert outbound links.

nistormihai:
The [#322](https://github.com/superdesk/liveblog/pull/322) is a quick fix and a fix for posts already added. (legacy), this is the proper fix. Also will not disallow theme developer to use other targets, caz they can use filters also.
In the future we can drop #322 PR.

vied12:
We can't drop #322 because we have to continue to support old posts already sent.
That's why I prefer to drop #323, to avoid duplicated feature. 
Then we keep the saved posts with tags as generic as possible, without extra parameters.  Like that, this is the responsibility of the theme to decide how to render tags. With `target="_parent"` or  `target="_blank"` depending the need.

PR #324, opened 2016-02-10
LBSD-774 autoupdate blog settings

vladnicoara:
As an editor, I want all changes to the blog settings (blog title,
description, image) to be auto-updated in the embed without need to
refresh the page (embed) manually.

### DIFF
--- a/server/liveblog/themes/themes_assets/classic/main.js
+++ b/server/liveblog/themes/themes_assets/classic/main.js
@@ -21,16 +21,7 @@
 
         function retrieveBlogSettings() {
             blogsService.get({}, function(blog) {
-                if (blog.title !== vm.blog.title) {
-                    vm.blog.title = blog.title;
-                }
-                if (blog.description !== vm.blog.description) {
-                    vm.blog.description = blog.description;
-                }
-                if (blog.picture_url !== vm.blog.picture_url) {
-                    vm.blog.picture_url = blog.picture_url;
-                    vm.blog.picture_srcset = blog.picture_srcset;
-                }
+                angular.extend(vm.blog, blog)
             });
         }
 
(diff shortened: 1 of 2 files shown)


server/liveblog/themes/themes_assets/classic/main.js
```diff
@@ -75,6 +91,7 @@
         .then(function() {
             vm.permalinkScroll();
             $interval(retrieveUpdate, UPDATE_EVERY);
+            $interval(retrieveBlogSettings, UPDATE_EVERY);
```
vied12: I would at least double the interval for `retrieveBlogSettings`

server/liveblog/themes/themes_assets/classic/main.js
```diff
@@ -19,6 +19,21 @@
...
 
+        function retrieveBlogSettings() {
+            blogsService.get({}, function(blog) {
+                if (blog.title !== vm.blog.title) {
+                    vm.blog.title = blog.title;
+                }
+                if (blog.description !== vm.blog.description) {
+                    vm.blog.description = blog.description;
+                }
+                //@TODO add the renditions updates once they are provided by the backend
```
vied12: is that something you plan to fix later, or did you forget to do it here ?

server/liveblog/themes/themes_assets/classic/main.js
```diff
@@ -19,6 +19,21 @@
...
+            blogsService.get({}, function(blog) {
+                if (blog.title !== vm.blog.title) {
+                    vm.blog.title = blog.title;
+                }
+                if (blog.description !== vm.blog.description) {
+                    vm.blog.description = blog.description;
+                }
+                if (blog.picture_url !== vm.blog.picture_url) {
+                    vm.blog.picture_url = blog.picture_url;
+                    vm.blog.picture_srcset = blog.picture_srcset;
```
nistormihai: angular.extend(vm.blog, blog) will do all this checks.
assigning only known properties can cause problems in the future if other properties will be added or will be counted, without knowing what is wrong.

ex: adding a update date for blog in template, you will expect that the update is ( updated ).

nistormihai:
In rest LGTM

PR #330, opened 2016-02-19
LBSD-824 highlights in timeline

ancafarcas:


vied12:
:+1: and e2e tests would be useful here

client/app/scripts/liveblog-edit/directives.js
```diff
@@ -253,6 +256,26 @@ define([
                                     notify.error(gettext('Something went wrong. Please try again later'));
                                 });
                             },
+                            highlightPost: function(post) {
+                                scope.clearReorder();
+                                if (post.highlight == false) {
```
nistormihai: please remove this is is not needed.
you can just use two messages base on the highlight.
ex: notify.info( post.highlight? gettext('Post was highlighted') ...

client/app/scripts/liveblog-edit/directives.js
```diff
@@ -253,6 +256,26 @@ define([
                                     notify.error(gettext('Something went wrong. Please try again later'));
                                 });
                             },
+                            highlightPost: function(post) {
+                                scope.clearReorder();
```
nistormihai: this is not needed, reorder isn't affected by highlight.

nistormihai:
LGTM great work :+1:

PR #332, opened 2016-02-25
chore(server: scripts: fig_wrapper): read db address from env vars

actionless:


actionless:
bamboo tests are passing

PR #338, opened 2016-02-29
fix: blog image removal and updates.

nistormihai:


client/app/scripts/liveblog-edit/module.js
```diff
@@ -312,7 +312,6 @@ define([
                 modal.confirm(gettext('Are you sure you want to remove the blog image?')).then(function() {
                     deregisterPreventer();
                     vm.newBlog.picture_url = null;
-                    vm.newBlog.picture = null;
```
vied12: Why stopping to remove the picture reference ?
the field still exists in `blog.py`
`'picture': Resource.rel('archive', embeddable=True, nullable=True, type='string'),`

nistormihai: If we update with null, then is not possible to pull blogs with `picture` object via embed.
And a PUT method to actually remove that property it will cause different issues ( I think you had those issues ).

Also this solution is actually used form superdesk in `avatar` removal process.

PR #344, opened 2016-03-08
Feature/sticky posts

vladnicoara:


client/app/scripts/liveblog-edit/directives.js
```diff
@@ -41,8 +43,9 @@ define([
                         originalOrder: 0,
                         pagesManager: new PagesManager($scope.lbPostsBlogId,
                                                        $scope.lbPostsStatus,
-                                                       10,
-                                                       $scope.lbPostsOrderBy || 'editorial'),
+                                                       $scope.lbSticky === 'true'? 100: 10,
```
vied12: `true` as a string ? Can't we use boolean ?
this makes not a lot of sense to me. Why number of post depends of the sticky posts ?
Could you add comments here to explain why ?
Otherwise we will forget and it could be removed later...

client/app/scripts/liveblog-edit/module.js
```diff
@@ -82,6 +82,7 @@ define([
             preview: false,
             actionPending: false,
             actionDisabled: true,
+            post: {sticky: false},
```
vied12: we already have a `currentPost`. Could we find a less confusing name ?

client/app/scripts/liveblog-edit/module.js
```diff
@@ -135,15 +141,24 @@ define([
...
+                var needsRemoval = $scope.currentPost && $scope.currentPost.sticky !== $scope.post.sticky;
                 postsService.savePost(blog._id,
                     $scope.currentPost,
                     getItemsFromEditor(),
-                    {post_status: 'open'}
+                    {post_status: 'open', sticky: $scope.post.sticky}
                 ).then(function(post) {
                     notify.pop();
                     notify.info(gettext('Post saved'));
+                    if (needsRemoval) {
```
vied12: Why do we remove in the `publish` method ? :)
I think we already deal with deletion somewheere for update. Like when we publish a draft, we remove it from from the draft list. It would be awesome to trigger that at the same place

client/app/scripts/liveblog-edit/views/main.html
```diff
@@ -177,12 +184,50 @@ <h3 class="page-nav-title">{{ blog.title }}</h3>
...
+                        lb-sticky='true'
+                        lb-posts-allow-unpublishing="blogSecurityService.canPublishAPost"
+                        lb-posts-allow-reordering="false"
+                        lb-posts-allow-editing="blogSecurityService.canPublishAPost"
+                        lb-posts-allow-deleting="blogSecurityService.canPublishAPost"
+                        lb-posts-on-post-selected="openPostInEditor"
+                        >
+                    </div>
+
+                    <!-- <div ng-if="!blogEdit.timelineStickyInstance.isPostsEmpty()" style="cursor:pointer"
```
vied12: Please remove comments

server/liveblog/posts/posts.py
```diff
@@ -156,7 +160,7 @@ def on_create(self, docs):
             doc['type'] = 'composite'
             doc['order'] = self.get_next_order_sequence(doc.get('blog'))
             # if you publish a post directly which is not a draft it will have a published_date assigned
-            if doc['post_status'] in ('open', 'sticky'):
+            if doc['post_status'] in ('open'):
```
vied12: no need to instantiate a tuple to compare strings here

server/liveblog/themes/themes_assets/classic/main.js
```diff
@@ -5,20 +5,30 @@
     function TimelineCtrl($interval, PagesManager, blogsService, config, $anchorScroll, $timeout, Permalink, transformBlog) {
 
         var POSTS_PER_PAGE = config.settings.postsPerPage;
+        var STICKY_POSTS_PER_PAGE = 100;
```
vied12: wow. A lot ! How do we handle pagination with sticky post btw ?

server/liveblog/themes/themes_assets/classic/main.js
```diff
@@ -5,20 +5,30 @@
...
         var PERMALINK_DELIMITER = config.settings.permalinkDelimiter || '?';
         var DEFAULT_ORDER = config.settings.postOrder; // newest_first, oldest_first or editorial
         var UPDATE_MANUALLY = config.settings.loadNewPostsManually;
         var UPDATE_EVERY = 10*1000; // retrieve update interval in millisecond
         var vm = this;
-        var pagesManager = new PagesManager(POSTS_PER_PAGE, DEFAULT_ORDER),
+        var pagesManager = new PagesManager(POSTS_PER_PAGE, DEFAULT_ORDER, false),
```
vied12: these PageManager constructors (in theme and editor) start to be hard to be understandable.
Shouldn't we use a single object as parameter, then we can name the properties ?

nistormihai: and object as parameter should be fine, I agree with @vied12.
But I think we should do it in another task, a much more technical debed refactor.

server/liveblog/themes/themes_assets/classic/main.js
nistormihai: sticky are not handled by pagination.

vied12: I meant API pagination. What to do if there are more than 100 sticky posts for instance?

nistormihai:
LGTM

PR #349, opened 2016-03-15
LBSD-843 blogs are redeployed when theme settings change

ancafarcas:


vied12:
:+1:
We still have to understand why requesting blogs filtered by theme doesn't work as it was implemented before

server/liveblog/themes/themes.py
```diff
@@ -239,12 +238,20 @@ def publish_related_blogs(self, theme):
...
+        blogs = get_resource_service('blogs').get(req=None, lookup={})
+        # get all the children for the theme that we modify the settings for
+        theme_children = self.get_children(theme.get('name'))
         for blog in blogs:
-            publish_blog_embed_on_s3.delay(str(blog['_id']))
+            blog_pref = blog.get('blog_preferences')
+            if theme_children:
+                # if a blog has associated the theme that is a  child of the one
+                # for which we modify the settings, we redeploy the blog on s3
+                for child in theme_children:
```
vied12: this can be optimized by leaving the loop when the wanted blog is found

server/liveblog/themes/themes.py
```diff
@@ -239,12 +238,20 @@ def publish_related_blogs(self, theme):
...
+                    if blog_pref['theme'] == child:
+                        publish_blog_embed_on_s3.delay(str(blog['_id']))
+            if blog_pref['theme'] == theme['name']:
```
vied12: I would put this check before the children loop. Then if it is true, it can avoid the loop

server/liveblog/themes/themes.py
```diff
@@ -239,12 +238,20 @@ def publish_related_blogs(self, theme):
         terms = []
         for t in self.get_children(theme['name']) + [theme['name']]:
             terms.append({'term': {'blog_preferences.theme': t}})
```
vied12: This is not needed any more I guess. But you can comment it just after the `FIXME`

server/liveblog/themes/themes.py
ancafarcas: The loop is anyway avoided if the theme doesn't have children. And if it does, then it should go through that loop.

server/liveblog/themes/themes.py
ancafarcas: I agree

server/liveblog/themes/themes.py
ancafarcas: Right, will do :)

server/liveblog/themes/themes.py
vied12: if `blog_pref['theme'] == theme['name']` is true, there is no need to loop over children to check if their name match with `blog_pref['theme']`. Because we already know the matching theme :)

ancafarcas: ok, fair enough

PR #355, opened 2016-03-22
LBSD-828 Disable draft button if not owner

vladnicoara:
Draft button should be disabled or hidden when a user opens a
contribution created by another user


ancafarcas:
LGTM :+1:

PR #370, opened 2016-04-12
Fix/ui fixes

darconny:


client/app/scripts/liveblog-edit/views/main.html
```diff
@@ -186,8 +186,10 @@ <h3 class="page-nav-title">{{ blog.title }}</h3>
                 </div>
                 <div class="header--live">
                     <a href="{{ publicUrl }}" target="_blank" ng-disabled="!publicUrl" translate>Live</a>
-                    <button class="btn pull-right" ng-click="filterHighlight()" translate>All</button>
-                    <button class="btn pull-right" ng-click="filterHighlight(true)" translate>Highlights</button>
+                </div>
+                <div class="header--filters">                                        
+                    <button class="btn btn-xs active" ng-click="filterHighlight(); filter.isSelected ? (filter.isSelected = !filter.isSelected) : ''" ng-class="{true: 'active', false: ''}[!filter.isSelected]" tooltip-placement="bottom" tooltip="Show all" translate>All</button>
```
nistormihai: this expression is a bit awkward `filter.isSelected ? (filter.isSelected = !filter.isSelected) : ''`

darconny: I used this as a basic toggle code, it is working fine. If you could rewrite this later on and make it better or outside of html, please do so.

nistormihai: I think I will do it in this PR so I won't forget,
but the expression does something else then toggle, if isSelected is true it will make it false and if it is false it will make it empty
a plain toggle is just `filter.isSelected = !filter.isSelected`

darconny: Sure, just keep in mind this is like a radio toggle between two stated - either All (default) or Highlights is selected. Thanks.

PR #374, opened 2016-04-18
feature: remove blog from media.

nistormihai:
LBSD-831


ancafarcas:
Looks good but the behave tests and flake8 are failling so you should verify them

ioanpocol:
:+1:

PR #383, opened 2016-05-06
LBSD-795 "Blog Edit" text leads to "Blog List"

vladnicoara:


ancafarcas:
LGTM

client/app/styles/less/liveblog.less
```diff
@@ -153,3 +153,9 @@
 .modal-body.split-popup>.sidebar>ul>li:nth-of-type(3) {
     display: none;
 }
+
+//disable main navigation link (does nothing for liveblog)
+#top-menu a.main-link {
+    cursor: default;
+    pointer-events: none;
```
nistormihai: this is funny, so it wasn't possible from superdesk or like that?

vladnicoara: I did not see an obvious way, setting that to a falsy value would have hidden the whole 'a' tag, and we need to show the text for a "breadcrumbs like" effect

PR #384, opened 2016-05-07
LBSD-895 All editors in the blog team should be able to edit blog set…

vladnicoara:
…tings

### DIFF
--- a/client/app/scripts/liveblog-security.service.js
+++ b/client/app/scripts/liveblog-security.service.js
@@ -32,7 +32,7 @@ angular.module('liveblog.security', [])
             return $rootScope.currentUser._id === archive.original_creator || canPublishAPost();
         }
         function canAccessSettings(archive) {
-            return canCreateABlog() && isUserOwnerOrAdmin(archive);
+            return canCreateABlog() && (isUserOwnerOrAdmin(archive) || isMemberOfBlog(archive));
         }
         function goToSettings() {
             var def = $q.defer();


client/app/scripts/liveblog-security.service.js
```diff
@@ -32,7 +32,7 @@ angular.module('liveblog.security', [])
             return $rootScope.currentUser._id === archive.original_creator || canPublishAPost();
         }
         function canAccessSettings(archive) {
-            return canCreateABlog() && isUserOwnerOrAdmin(archive);
+            return canCreateABlog() && (isUserOwnerOrAdmin(archive) || isMemberOfBlog(archive));
```
ancafarcas: after the change you made here, all the members of the blogs will have the rights to edit the settings, not only the editors, right?

nistormihai: Anca is right it should be only for '"editor" role that are in the blog team'
this grands right for all the users not only for "editor"s.

vladnicoara: it must have canCreateABlog right as well and that means admins and editors. Contributors do not have the rights to create blog. Tested the thing. You can check how it works on the branch

ancafarcas: You're right! Despite of what is understandable from the code, if you log in with contributor account you don't have the rights to create blogs. Tested on the branch.
So :+1:

PR #400, opened 2016-06-01
Fix/quote item dot fix

darconny:


ancafarcas:
@nistormihai @darconny I verified the branch and the credit appears in the right hand side for the emebed plus when you upload an embed the author (the credit) appears twice.

nistormihai:
LGTM test need to pass

PR #406, opened 2016-06-16
LBSD-938: Name is not shown if name format is not "full name" in them…

vladnicoara:
…e settings


nistormihai:
It needs build.
This needs to be PR first in the https://github.com/liveblog/lb-theme-classic repo
and then merge in this repo.

Other then that LGTM

PR #408, opened 2016-06-21
Fix/comments

vladnicoara:
Reader comments: Name is not shown if name format is not "full name" in theme settings


nistormihai:
LGTM

PR #417, opened 2016-07-01
Feature/iframe resizer

nistormihai:
* added setting in blog setting to make this height fixed(normal)/autoresize height switch.
* added `parent-iframe.js` script/marker to make a theme resize on height.
* use `https` in themes where is avalible.
* `classic` theme build


vladnicoara:
LGTM

PR #431, opened 2016-07-29
Fix/item text markup

nistormihai:
* added `fixMarkup` filter in angular theme.
* applied that filter into classic theme

**PR needs also merged**:

* https://github.com/liveblog/lb-theme-angular/pull/5
* https://github.com/liveblog/lb-theme-classic/pull/10


vladnicoara:
LGTM

PR #432, opened 2016-08-01
Fix/user avatar setsrc

nistormihai:
* if there is thumbnail use that
* if not pick the smallest one
* use srcset also for user avatar.
LBSD-955

PR needs to be merged as well:
* https://github.com/liveblog/lb-theme-angular/pull/6
* https://github.com/liveblog/lb-theme-classic/pull/11


ancafarcas:
LGTM

vladnicoara:
LGTM

PR #437, opened 2016-08-09
LBSD-970 json file for elastic mapping

ancafarcas:


server/liveblog/items/items.py
```diff
@@ -77,6 +78,16 @@ def on_create(self, docs):
         for doc in docs:
             update_dates_for(doc)
             doc['original_creator'] = str(get_user().get('_id'))
+            if doc.get('item_type'):
```
nistormihai: can we have just one if here ? the imbrication is not so nice
`if doc.get('item_type','') == 'embed': `

server/requirements.txt
```diff
@@ -4,5 +4,5 @@ Flask-Cors==2.1.0
 # Flask-S3 with support for python3
 git+git://github.com/joehand/flask-s3@8fe9f27ebb5251b7cf53a6684595d41dc4af2fbf#egg=Flask-S3==0.1.7
 # Superdesk
--e git+git://github.com/liveblog/superdesk-core@0bf96c9bfc9b1ce7e8049a5998fb3ad3abfdf3cc#egg=Superdesk-Core==1.05-test
+-e git+git://github.com/liveblog/superdesk-core@90ce33bae45873a5ec567df01571c8af231c1eab#egg=Superdesk-Core==1.06-test
```
nistormihai: this commit it is in superdesk/superdesk-core.

PR #439, opened 2016-08-11
Fix/keep embed responsive url

nistormihai:
* moved `parent-iframe.js` into angular from classic theme.
* when publishing on s3 `parent-iframe.js` don't use version.
* versioning for this file is based on time 15mins period cache.


vladnicoara:
LGTM

PR #451, opened 2016-08-24
LBSD-1001 Fix for instagram embed codes in blog edit view

vladnicoara:


nistormihai:
now  that is it.
LGTM

PR #455, opened 2016-08-31
Fix/test update

ancafarcas:


server/liveblog/blogs/blogs.py
```diff
@@ -32,7 +32,7 @@
 
 blogs_schema = {
     'title': metadata_schema['headline'],
-    'description': metadata_schema['description'],
+    'description': metadata_schema['description_text'],
```
nistormihai: How this will change existing blogs?

server/liveblog/blogs/blogs.py
```diff
@@ -94,8 +94,11 @@ def send_email_to_added_members(blog, recipients, origin):
     recipients_email = []
     for user in recipients:
         # if user want to receive email notification, we add him as recipient
-        if prefs_service.email_notification_is_enabled(user_id=user['user']):
-            user_doc = get_resource_service('users').find_one(req=None, _id=user['user'])
+        if prefs_service.email_notification_is_enabled(user_id=user):
+            if isinstance(user, ObjectId):
```
nistormihai: can we have a single method of doing this? caz is used also below

server/liveblog/blogs/blogs.py
```diff
@@ -194,7 +203,10 @@ def on_updated(self, updates, original):
         recipients = []
         for user in members:
```
nistormihai: and here

server/ws.py
```diff
@@ -10,7 +10,7 @@
 # at https://www.sourcefabric.org/superdesk/license
 
 
-from settings import WS_HOST, WS_PORT, LOG_SERVER_ADDRESS, LOG_SERVER_PORT
+from settings import WS_HOST, WS_PORT, LOG_SERVER_ADDRESS, LOG_SERVER_PORT, BROKER_URL
```
nistormihai: what is a broker?

server/apps/prepopulate/data_initialization/users.json
```diff
@@ -0,0 +1,15 @@
...
+    {
+        "_id": "5788bc55ea8e157ca2ab7629",
+        "display_name": "admin admin",
+        "email": "admin+123@superdesk.org",
+        "first_name": "admin",
+        "is_active": true,
+        "last_name": "admin",
+        "needs_activation": false,
+        "password": "$2a$12$mJVAxOc/MgFla8RkxJ87RuxvcoXEjc75nngkS0GNb52SNKxRcGDTy",
+        "role": "54e650f31024542de2649bea",
```
ancafarcas: Honestly, this `users.json` file was copied from superdesk, so I have no idea.

server/liveblog/blogs/blogs.py
ancafarcas: It won't change the old blogs. In the `blogs` endpoint the name remains `description`

server/ws.py
ancafarcas: I'm not really sure. Again...this is something took for granted from superdesk because otherwise I was obtaining an error. Anyway, it something that celery needs.

server/liveblog/blogs/blogs.py
ancafarcas: I remember I tried to put it into a single method but for some reasons it was keep failing and I let it aside. I could try again right now.

server/ws.py
nistormihai: is required to be set?

nistormihai:
other than the comments LGTM

server/liveblog/blogs/blogs.py
ancafarcas: Our database field doesn't change. It is only pointing now to `description_text` which is defined in the superdesk schema

server/ws.py
ancafarcas: yes

server/liveblog/blogs/blogs.py
ancafarcas: I just remembered why I let aside creating a method for this check-up since it would have been to complicated. In the first case where I used `isinstance (user, ObjectId)` I then needed to get the whole user document after the verification in order to be able to extract its email. In the second case, I only needed to do the simple check-up.

PR #465, opened 2016-09-20
Feature/server optimization

nistormihai:


server/liveblog/client_modules/client_modules.py
```diff
@@ -65,7 +70,16 @@ class ClientBlogsResource(BlogsResource):
 
 
 class ClientBlogsService(BaseService):
-    pass
+    def find_one(self, req, **lookup):
+        if req is None:
+            req = ParsedRequest()
+        cache_key = 'lb_ClientBlogService_get_%s' % (hash(frozenset(req.__dict__.items())))
```
djbrd-sourcefabric: Definition of key should be in a separate method, not copied, so that it only has to be changed once if necessary

server/liveblog/client_modules/client_modules.py
```diff
@@ -160,8 +174,8 @@ class ClientBlogPostsService(BlogPostsService):
     def get(self, req, lookup):
         cache_key = 'lb_ClientBlogPostsService_get_%s' % (hash(frozenset(req.__dict__.items())))
         blog_id = lookup.get('blog_id')
-        docs = app.blog_cache.get(blog_id, cache_key)
+        docs = app.liveblog_cache.get(blog_id, cache_key)
```
djbrd-sourcefabric: 5 lines here are the same as 5 lines above. Please put in a separate method. Also, is it truly impossible in the code to update the document without updating the cache?

server/liveblog/common.py
```diff
@@ -23,29 +26,66 @@ def check_comment_length(text):
...
+        blog_cache_key = '{}__{}__{}_version'.format(app.config['INSTANCE_ID'],
+                                                     blog, self.main_key)
+        blog_version = self.cache.get(blog_cache_key)
+        if not blog_version:
+            blog_version = 1
+            self.cache.set(blog_cache_key, blog_version)
+        if invalidate:
+            # @TODO: check performance before uncoment
+            # self.__remove_cache_keys(blog, blog_version=blog_version)
+            if not blog_version:
```
djbrd-sourcefabric: Why are these 2 lines here, when they are also just above (line 44)

server/liveblog/common.py
```diff
@@ -23,29 +26,66 @@ def check_comment_length(text):
...
         self.cache = cache
+        self.main_key = key
 
     def __get_blog_version(self, blog, invalidate=False):
         '''
         Return the blog cache version.
         If invalidate is true, the version will be incremented
         '''
-        return update_key('%s__version' % (blog), flag=invalidate)
+        blog_cache_key = '{}__{}__{}_version'.format(app.config['INSTANCE_ID'],
```
djbrd-sourcefabric: Again, please put generation of key in a method

server/liveblog/common.py
```diff
@@ -23,29 +26,66 @@ def check_comment_length(text):
...
+        blog_version = self.cache.get(blog_cache_key)
+        if not blog_version:
+            blog_version = 1
+            self.cache.set(blog_cache_key, blog_version)
+        if invalidate:
+            # @TODO: check performance before uncoment
+            # self.__remove_cache_keys(blog, blog_version=blog_version)
+            if not blog_version:
+                blog_version = 1
+            blog_version = (blog_version + 1) % 256
```
djbrd-sourcefabric: Why 256?

server/liveblog/client_modules/client_modules.py
djbrd-sourcefabric: And I think that the generation of the key should be in the liveblog_cache class

server/liveblog/posts/posts.py
```diff
@@ -174,7 +179,7 @@ def on_create(self, docs):
             if doc['post_status'] == 'open':
                 doc['published_date'] = utcnow()
                 doc['content_updated_date'] = doc['published_date']
-                doc['publisher'] = getattr(flask.g, 'user', None)
+                doc['publisher'] = {'_id': getattr(flask.g, 'user', {}).get('_id', None)}
```
djbrd-sourcefabric: Please put in a _get_publisher method

server/app.py
```diff
@@ -57,9 +60,16 @@ def get_app(config=None):
     ])
     app.jinja_loader = custom_loader
 
-    # cache
-    app.cache = Cache(app, config={'CACHE_TYPE': 'simple'})
-    app.blog_cache = BlogCache(cache=app.cache)
+    # cache definition
+    # right now mostly used on client_modules
+    if config['MEMCACHE_URI']:
```
djbrd-sourcefabric: Why use memcache and not redis, which is already a dependency?

PR #473, opened 2016-09-28
feature: if external media isn't supported, compose the local url.

nistormihai:
So that the `Live` link shows even if s3 is not used, in production enviroment.


ancafarcas:
LGTM :+1:

PR #474, opened 2016-09-28
[WIP] LBSD-1028 - Upgrading to modern superdesk-client-core

m0g:
Beware, this is very much a work in progress


.gitignore
```diff
@@ -67,3 +67,5 @@ dump.rdb
 *.swn
 *.swm
 .vimrc
+
+client/config.js
```
nistormihai: in this configuration we will put the embedly-key?

client/package.json
```diff
@@ -12,6 +12,9 @@
     }
   ],
   "devDependencies": {
+    "babel-core": "^6.10.4",
+    "babel-loader": "^6.2.4",
+    "babel-preset-es2015": "^6.9.0",
     "btoa": "^1.1.2",
     "grunt": "^0.4.2",
```
nistormihai: what do you say, how big of an effort it is to use refactor grunt with gulp?

client/webpack.config.js
```diff
@@ -0,0 +1,176 @@
...
+            // if true, the editor will not have a toolbar
+            disableEditorToolbar: grunt.option('disableEditorToolbar')
+        },
+
+        // default timezone for the app
+        defaultTimezone: grunt.option('defaultTimezone') || 'Europe/London',
+
+        // model date and time formats
+        model: {
+            dateformat: 'DD/MM/YYYY',
```
nistormihai: if this is momentjs format, can we use `L` local format date and `LT` for time?

.gitignore
m0g: Embedly key, server url, google analytics key and so on

client/package.json
m0g: To replace grunt with gulp you mean?

I personally prefer gulp, but since ```superdesk-client-core``` is relying on grunt, I think we should keep on using it.

nistormihai: this question is just to estimate the effort

.gitignore
nistormihai: and how the environment keys are set in this process?
I want to allow travis to work on forks but for that I still need the `env` variables.

client/package.json
m0g: Yes, but to be fair, what would be the benefit of switching to gulp?

nistormihai: we use `gulp` in theme management system, so unification of software could be a plus.

nistormihai:
Please remake this PR on devel branch.

m0g:
@nistormihai why should the PR point to the devel branch?

m0g:
https://github.com/liveblog/liveblog/pull/483

PR #478, opened 2016-09-29
Fix/keep authorship

nistormihai:


server/liveblog/posts/posts.py
```diff
@@ -164,7 +164,7 @@ def on_create(self, docs):
             # if you publish a post directly which is not a draft it will have a published_date assigned
             if doc['post_status'] == 'open':
                 doc['published_date'] = utcnow()
-                doc['content_updated_date'] = utcnow()
+                doc['content_updated_date'] = doc['published_date']
```
ancafarcas: Do you want the updated_date here to be the same as published_date?

server/liveblog/posts/posts.py
```diff
@@ -218,6 +218,9 @@ def on_update(self, updates, original):
             # if you publish a post it will save a published date and register who did it
             updates['published_date'] = utcnow()
             updates['publisher'] = getattr(flask.g, 'user', None)
+            # if you publish a post and hasn't `content_updated_date` add it.
+            if not updates.get('content_updated_date', False):
+                updates['content_updated_date'] = updates['published_date']
```
ancafarcas: the same question here: why do we need the updated_date to be the same as publihed_date?

server/liveblog/posts/posts.py
nistormihai: yes, it wasn't that clear that happend in the previous code.

server/liveblog/posts/posts.py
nistormihai: we need to have the `content_updated_date` the same as publish because the content is changed when publishing.
ex: make a post not open, edit that post until this time it doesn't have a `content_updated_date` if nothing is changed in the post the `content_updated_date` will not be compiled, so it will stay `undefined`. by this line we assure that there is a `content_updated_date` on a post.

PR #481, opened 2016-09-30
LBSD-1029 Prevent form submit on Enter key

vladnicoara:
When pressing the Enter key in the title field, an unwanted "Add to dictionary" dialog opened in the blog description area. Now the Enter key forwards the form to the next page if the condition is met.


nistormihai:
LGTM

PR #497, opened 2016-10-28
LBSD-1091:  Add api_url to consumer resource

mscam:
Add api_url to consumer resource. We need to add api_url to consumer resource. This is required to get the syndication webhook working. We no longer need to define a target_url for syndicate webhook.


m0g:
LGTM

djbrd-sourcefabric:
:+1:

PR #503, opened 2016-11-07
LBSD-1085 Contacts may be added to both consumers and producers

djbrd-sourcefabric:
n contacts may be added to consumer and producer on the client. Contacts are persisted to the consumer whereas they weren't before


m0g:
:+1:

mscam:
:+1:

PR #509, opened 2016-11-11
Added auto_publish flag to in_syndication

djbrd-sourcefabric:
Added auto_publish flag to in_syndication. Status of post is set accordingly in webhook.


m0g:
To avoid confusion, can we just call the field `autopublish` ?

m0g:
All right

mscam:
:+1:

PR #513, opened 2016-11-15
fix: `ELASTICSEARCH_FORCE_REFRESH` True.

nistormihai:


m0g:
All right,

so what is this for?

nistormihai:
So we have this notification issue, where the push event is sent before the index are refreshed.
It seems like it is not getting the actions performed.
in elastic eve_elastic is set the true by default but somehow or somewhere it is changed and it is not working.

https://github.com/petrjasek/eve-elastic/blob/master/eve_elastic/elastic.py#L184

m0g:
All right,

:+1:

PR #522, opened 2016-11-22
Autopublish fix.

mscam:
New and old posts are now posted by using post_status "open" if auto_publish is set to True and with post_status "submitted" if auto_publish is set to False

### DIFF
--- a/server/liveblog/syndication/syndication.py
+++ b/server/liveblog/syndication/syndication.py
@@ -286,7 +286,12 @@ def _create_blog_post(old_post, items, in_syndication, post_status=None):
         })
 
     if not post_status:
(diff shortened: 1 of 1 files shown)


server/liveblog/syndication/syndication.py
```diff
@@ -286,7 +286,11 @@ def _create_blog_post(old_post, items, in_syndication, post_status=None):
         })
 
     if not post_status:
-        post_status = 'open' if in_syndication.get('auto_publish', False) else 'submitted'
+        auto_publish = in_syndication.get('auto_publish', False)
+        if auto_publish:
+            post_status = 'open'
+        else:
+            post_status = 'submitted'
```
m0g: So I suppose this is just for style?

m0g:
:+1: all good

PR #523, opened 2016-11-22
Fix/keep authorship all

nistormihai:
keep authorship for post submitted when they change to open even if they are changed by others.


mscam:
:+1:

PR #528, opened 2016-11-25
LBSD-1141: Check for duplicate posts on syndication webhook request.

mscam:
- Check for duplicate posts on syndication webhook request (only for new posts)
- New formatting for producer_post_id post field

### DIFF
--- a/server/liveblog/syndication/tasks.py
+++ b/server/liveblog/syndication/tasks.py
@@ -13,11 +13,14 @@ logger = logging.getLogger('superdesk')
 @celery.task(bind=True)
 def send_post_to_consumer(self, syndication_out, producer_post, action='created'):
     """Send blog post updates to consumers webhook."""
-    from .utils import extract_post_items
+    from .utils import extract_post_items, extract_producer_post_data
     consumers = get_resource_service('consumers')
     items = extract_post_items(producer_post)
     try:
-        consumers.send_post(syndication_out, {'items': items, 'producer_post': producer_post}, action)
+        consumers.send_post(syndication_out, {
+            'items': items,
+            'producer_post': extract_producer_post_data(producer_post)
+        }, action)
     except APIConnectionError as e:
         raise self.retry(exc=e, max_retries=SYNDICATION_CELERY_MAX_RETRIES, countdown=SYNDICATION_CELERY_COUNTDOWN)
 
@@ -25,7 +28,7 @@ def send_post_to_consumer(self, syndication_out, producer_post, action='created'
 @celery.task(bind=True)
 def send_posts_to_consumer(self, syndication_out, action='created', limit=25):
     """Send latest blog post updates to consumers webhook."""
-    from .utils import extract_post_items
+    from .utils import extract_post_items, extract_producer_post_data
     consumers = get_resource_service('consumers')
     blog_id = syndication_out['blog_id']
     posts_service = get_resource_service('posts')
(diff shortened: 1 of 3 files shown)


m0g:
:+1:

server/liveblog/syndication/tasks.py
```diff
@@ -0,0 +1,57 @@
...
+from settings import SYNDICATION_CELERY_MAX_RETRIES, SYNDICATION_CELERY_COUNTDOWN
+
+
+logger = logging.getLogger('superdesk')
+
+
+@celery.task(bind=True)
+def send_post_to_consumer(self, syndication_out, producer_post, action='created'):
+    """Send blog post updates to consumers webhook."""
+    from .utils import extract_post_items, extract_producer_post_data
```
djbrd-sourcefabric: I think it would be better to do imports at the top of the file

server/liveblog/syndication/tasks.py
```diff
@@ -0,0 +1,57 @@
...
+    consumers = get_resource_service('consumers')
+    items = extract_post_items(producer_post)
+    try:
+        consumers.send_post(syndication_out, {
+            'items': items,
+            'producer_post': extract_producer_post_data(producer_post)
```
djbrd-sourcefabric: The call to the function should be before the call to send_post, I think, as it is with the items, just to make the code more readable and easier to debug

mscam: :+1:

server/liveblog/syndication/tasks.py
```diff
@@ -0,0 +1,57 @@
...
+    """Send latest blog post updates to consumers webhook."""
+    from .utils import extract_post_items, extract_producer_post_data
+    consumers = get_resource_service('consumers')
+    blog_id = syndication_out['blog_id']
+    posts_service = get_resource_service('posts')
+    posts = posts_service.find({'blog': blog_id}).limit(limit)
+    try:
+        for producer_post in posts:
+            producer_post_type = producer_post.get(ITEM_TYPE, '')
+            if producer_post_type == CONTENT_TYPE.COMPOSITE:
```
djbrd-sourcefabric: Bug here, I think. Because the posts service is returning all post types, and the filtering is done here, there is no guarantee that 25 will be sent to the consumer. In fact, I guess that a maximum of 13 will be sent.

server/liveblog/syndication/utils.py
```diff
@@ -96,3 +100,148 @@ def fetch_url(url, timeout=5):
             fd.write(chunk)
     fd.seek(0)
     return fd
+
+
+def extract_post_items(original_doc):
```
djbrd-sourcefabric: I actually think that this should be in syndication.py, as it extracts just the data which is required to send to the consumer via the webhook, and not the items themselves. Perhaps it should be renamed extract_post_items_data

server/liveblog/syndication/utils.py
```diff
@@ -96,3 +100,148 @@ def fetch_url(url, timeout=5):
...
+                data = {
+                    'text': item['text'],
+                    'item_type': item_type,
+                    'meta': item.get('meta', {})
+                }
+                items.append(data)
+    return items
+
+
+def get_html_from_image_data(renditions, **meta):
```
djbrd-sourcefabric: Prefix with underscore because it's only used locally

server/liveblog/syndication/utils.py
```diff
@@ -96,3 +100,148 @@ def fetch_url(url, timeout=5):
...
+        post_id
+    )
+
+
+def extract_producer_post_data(post, fields=('_id', '_updated', 'highlight', 'sticky', 'post_status')):
+    """Extract only useful data from original producer blog post."""
+    return {key: post[key] for key in fields}
+
+
+def create_syndicated_blog_post(producer_post, items, in_syndication, post_status=None):
```
djbrd-sourcefabric: Is there any time when post_status will not be None?

djbrd-sourcefabric:
:+1:

PR #531, opened 2016-11-29
LBSD-1148 - Publish & delete a post from incoming syndication

m0g:


mscam:
:+1:

client/app/scripts/liveblog-syndication/actions/incoming-syndication.js
```diff
@@ -27,6 +27,23 @@ liveblogSyndication
                             });
                         });
                     });
+                },
+                publish: function(post) {
+                    postsService.savePost(post.blog, post, undefined, {post_status: 'open'})
+                        .then(function(post) {
+                            Dispatcher.dispatch({
+                                type: 'ON_YANKED_POST',
```
djbrd-sourcefabric: Not sure what 'yanked' means here, or why it's being used for two quite different actions. For the sake of readability, I would have 2 different types (e.g. ON_SAVED_POST and ON_REMOVED_POST) which can both be handled by the same code in the listener

client/app/scripts/liveblog-syndication/directives/incoming-syndication.js
```diff
@@ -26,6 +26,20 @@ liveblogSyndication
...
 
+                    scope.publish = function(post) {
+                        IncomingSyndicationActions
+                            .publish(post);
+                    };
+
+                    scope.askRemove = function(post) {
+                        modal.confirm(gettext('Are you sure you want to delete the post?'))
+                            .then(function() {
+                                console.log('remove it yo!');
```
djbrd-sourcefabric: If you want to leave in debug messages, please make them meaningful

client/app/scripts/liveblog-syndication/actions/incoming-syndication.js
m0g: Both actions triggers the same reducer, which removes an entry from a the posts list. Hence the "yank"

djbrd-sourcefabric: To find out what 'yanked' means here (a word whose meaning isn't clear to me, a native speaker), I have to read the code. It means saved or removed. So, please change it to ON_SAVED_POST and ON_REMOVED_POST. You then just to have one more line of code, an extra case in the switch statement.

PR #534, opened 2016-11-30
LBSD-1157 - Fix css issue on resize consumers list

m0g:


djbrd-sourcefabric:
Surprised that there is no existing css class in bootstrap to fix the problem, but OK :+1:

PR #535, opened 2016-12-01
LBSD-1117: Syndication webhook: update posts

mscam:
Support for updating syndicated blog posts data using webhook (wip)

### DIFF
--- a/server/liveblog/syndication/syndication.py
+++ b/server/liveblog/syndication/syndication.py
@@ -193,7 +193,7 @@ def syndication_webhook():
 
         post_id = str(post['_id'])
         publisher = post.get('publisher')
-        if not publisher and method == 'PUT':
+        if not publisher:
             # Update only posts with publisher set to None
             posts_service.update(post_id, new_post, post)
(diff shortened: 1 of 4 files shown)


server/liveblog/syndication/syndication.py
```diff
@@ -174,14 +180,25 @@ def syndication_webhook():
...
+            return api_error('Post cannot be updated: syndicated post does not exist.', 404)
+
+        post_id = str(post['_id'])
+        publisher = post.get('publisher')
+        if not publisher and method == 'PUT':
+            # Update only posts with publisher set to None
+            posts_service.put(post['_id'], new_post)
+            return api_response({'post_id': post_id}, 200)
+        else:
+            return api_error('Post "{}" cannot be updated: publisher already exists.'.format(post_id), 409)
```
djbrd-sourcefabric: No sure that this should return an error, as it is an expected case and the source of the post doesn't have to respond to it

m0g:
:+1:

PR #538, opened 2016-12-02
Add multi-platform support to Vagrantfile

psolbach:
@m0g, this should play nice across platforms. One can avoid provisioning of all providers at once by altering `config.vm.box_url` conditionally, in effect just replacing urls for a shared box. I'd be happy to commit this on another branch if you like.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -2,9 +2,22 @@
 # vi: set ft=ruby :
 
 Vagrant.configure(2) do |config|
-    config.vm.box = "fgrehm/trusty64-lxc"
+    config.vm.box = "trusty64"
+    host = RbConfig::CONFIG['host_os']
 
-    config.vm.box_check_update = false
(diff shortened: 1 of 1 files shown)


m0g:
I'll try it on my personal laptop, the one that has the LXC setup.

Vagrantfile
```diff
@@ -13,8 +26,8 @@ Vagrant.configure(2) do |config|
     config.vm.network "forwarded_port", guest: 6379, host: 6379
     config.vm.network "forwarded_port", guest: 35729, host: 35729
 
-    config.vm.synced_folder ".", "/opt/liveblog"
+    config.vm.network :private_network, ip: '192.168.50.50' # Required for NFS to work, pick any local IP
```
m0g: Local networks don't work with LXC (as far as I can recall), so this should be inside the virtualbox block

PR #539, opened 2016-12-05
Lbsd 1172 - Display producer blog title in timeline

m0g:


mscam:
:+1:

client/app/scripts/liveblog-edit/posts.service.js
```diff
@@ -123,47 +123,64 @@ define([
...
+                                   (!post.hasComments) && (post.mainItem.item.item_type !== 'comment');
+                angular.forEach(post.items, function(val) {
+                    if (post.fullDetails) {
+                        _completeUser(val.item);
+                    }
+                });
+                _completeUser(post.mainItem.item);
+
+                if (post.syndication_in) {
+                    api.syndicationIn.getById(post.syndication_in).then(function(synd) {
```
djbrd-sourcefabric: Can we not cache the blog title somehow? Not sure if we need to worry about it, but this seems to be a pretty sub-optimal way of doing things. Alternative would be to do the preprocessing on the server, but you could just add an optimisation task in Jira for this.

m0g: Overall, a refactoring of post.service and pages manager would be required

PR #550, opened 2016-12-09
LBSD-1156: Delete in & out syndication on blog delete

mscam:
Delete in & out syndication on blog delete


mscam:
@m0g notification data: https://github.com/liveblog/liveblog/pull/550/files#diff-d45eb627b5002191692e04b1c03c26a3R106

PR #556, opened 2016-12-16
LBSD-1151: Allow consumer to have syndicated posts from a specific date

mscam:
- Added _start_date_ to _syndication_in_ & _syndication_out_ schema.
- Added PATCH support to _blogs_ and _syndication_ blueprints.
- Added behave test for _blogs_ syndicate blueprint.

### DIFF
--- a/server/liveblog/syndication/blogs.py
+++ b/server/liveblog/syndication/blogs.py
@@ -62,10 +62,10 @@ def _get_consumer_from_auth():
     return consumer
 
 
-def _create_blogs_syndicate(blog_id, consumer_blog_id):
+def _create_blogs_syndicate(blog_id, consumer_blog_id, start_date=None):
     # Get the blog to be syndicated - must be enabled for syndication
     blogs_service = get_resource_service('blogs')
-    blog = blogs_service.find_one(req=None, checkUser=False, _id=blog_id)
+    blog = blogs_service.find_one(req=None, checkUser=False, _id=blog_id) # TODO: Camel-case ist verboten!
     if blog is None:
         return api_error('No blog available for syndication with given id "{}".'.format(blog_id), 409)
     if not blog['syndication_enabled']:
@@ -76,11 +76,12 @@ def _create_blogs_syndicate(blog_id, consumer_blog_id):
     consumer_id = str(consumer['_id'])
     if out_service.is_syndicated(consumer_id, blog_id, consumer_blog_id):
         return api_error('Syndication already sent for blog "{}".'.format(blog_id), 409)
-    "{}"
+
     syndication_id = out_service.post([{
         'blog_id': blog_id,
         'consumer_id': consumer_id,
-        'consumer_blog_id': consumer_blog_id
+        'consumer_blog_id': consumer_blog_id,
+        'start_date': start_date
     }])[0]
     syndication = out_service.find_one(_id=syndication_id, req=None)
     return api_response({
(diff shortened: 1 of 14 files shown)


mscam:
Reminder: please remember to send the correct date format as specified by python-eve's _DATE_FORMAT_ configuration setting http://python-eve.org/config.html

server/liveblog/syndication/blogs.py
```diff
@@ -107,16 +129,26 @@ def _delete_blogs_syndicate(blog_id, consumer_blog_id):
...
 
-@blogs_blueprint.route('/api/syndication/blogs/<string:blog_id>/syndicate', methods=['POST', 'DELETE'])
+@blogs_blueprint.route('/api/syndication/blogs/<string:blog_id>/syndicate', methods=['POST', 'PATCH', 'DELETE'])
 def blogs_syndicate(blog_id):
-    consumer_blog_id = request.get_json().get('consumer_blog_id')
+    data = request.get_json()
+    start_date = data.get('start_date')
+    if start_date:
+        try:
+            start_date = str_to_date(start_date)
```
djbrd-sourcefabric: Might be nice to document somewhere, even if just in a comment, what is the expected format

server/liveblog/syndication/producer.py
```diff
@@ -189,6 +198,29 @@ def _create_producer_blogs_syndicate(producer_id, blog_id, consumer_blog_id, aut
...
+    try:
+        response = producers.syndicate(producer_id, blog_id, consumer_blog_id, start_date, update=True,
+                                       json_loads=False)
+    except ProducerAPIError as e:
+        return api_response(str(e), 500)
+    else:
+        if response.status_code == 200:
+            syndication_in = in_service.get_syndication(producer_id, blog_id, consumer_blog_id)
+            in_service.update(syndication_in['_id'], {'start_date': start_date}, syndication_in)
+            del syndication_in['blog_token']
```
djbrd-sourcefabric: This looks a little dangerous. Should be more difficult for the blog_token to be returned to the client. Do we actually need it to be exposed at all?

server/liveblog/syndication/blogs.py
```diff
@@ -76,20 +78,40 @@ def _create_blogs_syndicate(blog_id, consumer_blog_id):
...
         'blog_id': blog_id,
         'consumer_id': consumer_id,
-        'consumer_blog_id': consumer_blog_id
-    }])[0]
+        'consumer_blog_id': consumer_blog_id,
+        'start_date': start_date
+    }
+
+    syndication_id = out_service.post([doc])[0]
     syndication = out_service.find_one(_id=syndication_id, req=None)
```
mscam: unfortunately it's required as post() method is returning the document id, not the whole document, and we need the whole document to get and put the generated token in the response.

server/liveblog/syndication/producer.py
```diff
@@ -213,17 +245,28 @@ def _delete_producer_blogs_syndicate(producer_id, blog_id, consumer_blog_id):
...
+    if start_date:
+        try:
+            start_date = str_to_date(start_date)
+        except ValueError:
+            return api_error('start_date is not valid.', 400)
+
     if request.method == 'DELETE':
         return _delete_producer_blogs_syndicate(producer_id, blog_id, consumer_blog_id)
+    elif request.method == 'PATCH':
+        return _update_producer_blogs_syndicate(producer_id, blog_id, consumer_blog_id, start_date)
```
djbrd-sourcefabric: How is auto_publish value updated?

server/liveblog/syndication/syndication.py
```diff
@@ -33,11 +33,13 @@
...
     },
-    'last_delivered_post_id': Resource.rel('posts', embeddable=True, required=True,
-                                           nullable=True, type="objectid"),
     'token': {
         'type': 'string',
         'unique': True
+    },
+    'start_date': {
+        'type': 'datetime',
+        'default': None
```
djbrd-sourcefabric: Default to now?

server/liveblog/syndication/syndication.py
```diff
@@ -52,18 +54,24 @@ def _cursor(self, resource=None):
     def _get_blog(self, blog_id):
         return self._cursor('blogs').find_one({'_id': ObjectId(blog_id)})
 
-    def is_syndicated(self, consumer_id, producer_blog_id, consumer_blog_id):
+    def _lookup(self, consumer_id, producer_blog_id, consumer_blog_id):
```
djbrd-sourcefabric: If this is only used once, I wouldn't put it in a separate method

server/liveblog/syndication/syndication.py
```diff
@@ -143,25 +159,35 @@ class SyndicationOut(Resource):
...
+        'default': None
     }
 }
 
 
 class SyndicationInService(BaseService):
     notification_key = 'syndication_in'
 
-    def is_syndicated(self, producer_id, producer_blog_id, consumer_blog_id):
+    def _lookup(self, producer_id, producer_blog_id, consumer_blog_id):
```
djbrd-sourcefabric: Again, no need for a separate method here I think

server/liveblog/syndication/producer.py
mscam: blog_token should be returned to the client only when necessary, for security reasons, only by using SSL secure connection.

server/liveblog/syndication/producer.py
mscam: auto_publish can be updated by sending PATCH request to the blueprint url. This is not actually implemented by the javascript client, but it's available.

server/liveblog/syndication/syndication.py
mscam: Unfortunately python-eve doesn't set the default value I have selected (None) and set by default to datetime.datetime.now(). I am investigating about this issue.

PR #565, opened 2017-01-01
LBSD-1221 Split liveblog-edit module into manageable size files

vladnicoara:


nistormihai:
GOOD AND NEEDED.

mscam:
:+1:

PR #567, opened 2017-01-02
LBSD-1175: Change api_url to webhook_url

mscam:
- Change api_url to webhook_url on both backend and client app
- Updated behave tests and specs
- Fix for consumer list markup issue (username classname had different width)
- Updated httpsurl validator


djbrd-sourcefabric:
:+1:

m0g:
:+1:

PR #572, opened 2017-01-04
LBSD-1209: Change order of posts for autopublish on syndication

mscam:
Posts in syndicated blog on consumer side were ordered differently than on the producer's side when autopublish was enabled. The proposed bugfix changes the sorting of the blog posts, as unfortunately python-eve doesn't allow to update _updated or _inserted date values, that are always set to datetime.datetime.now() by the framework itself. We should have a different date field for blog posts sorting.


m0g:
:+1:

djbrd-sourcefabric:
:+1:

PR #586, opened 2017-01-11
LBSD-1238 Syndication switch

m0g:


mscam:
:+1:

client/tasks/options/template.js
```diff
@@ -18,6 +18,7 @@ module.exports = function(grunt) {
                 key: grunt.option('embedly-key') || process.env.EMBEDLY_KEY || ''
             },
             facebookAppId: grunt.option('facebook-appid') || process.env.FACEBOOK_APP_ID || '',
+            syndication: true,
```
nistormihai: can we set it to `false` by default for now?

client/app/index.html
```diff
@@ -252,6 +251,9 @@ <h3 translate>About</h3>
             'ngMessages'
         ];
 
+        if (config.syndication)
```
nistormihai: the best way to enable or disable it :+1:

client/tasks/options/template.js
m0g: oh sure, I'll do directly in https://github.com/liveblog/liveblog/pull/588

PR #590, opened 2017-01-12
Add download CSV functionality to analytics view

psolbach:
Transform analytics items to CSV, download via `link.download` as opposed to `window.open`. Should work with any HTML5 capable browser and IE, cf. http://caniuse.com/#search=download. Code is linted this time around.


m0g:
:+1:

PR #592, opened 2017-01-13
LBSD-1241: Prevent creation of consumer and producer with same webhook or api url

mscam:
- Fix unique validation for webhook_url and api_url
- Bugfix for token auth validation
- Behave tests to check unique url validation for api_url and webhook_url.


djbrd-sourcefabric:
:+1:

PR #595, opened 2017-01-24
LBSD-1245 Reset post type to default

vladnicoara:
Reset post type to default when saving, updating submitting and
publishing


nistormihai:
LGTM

PR #600, opened 2017-02-01
Fix/theme template requests

nistormihai:


m0g:
Is there a reason why we are included the build files in the git repo?

nistormihai:
That is good point, I wanted from a long time to change this behaviour but that is a bigger/another task.

nistormihai:
themes are dynamically added we need also a server side process to build files.

m0g:
All right

vladnicoara:
is there a JIRA ticket associated to this PR? it's not very obvious what you are trying to fix or why. Thanks

PR #607, opened 2017-02-08
LBSD-1260

darconny:


nistormihai:
only failed: 
`1) editor embed: add a youtube iframe in the editor`
but that is due to the fork issue.

PR #610, opened 2017-02-08
Fix/keep score

vladnicoara:
Load last scorecard data( in a copy) when opening a fresh scorecard to prevent having to enter all the scorers again

### DIFF
--- a/client/app/scripts/liveblog-edit/controllers/blog-edit.js
+++ b/client/app/scripts/liveblog-edit/controllers/blog-edit.js
@@ -70,12 +70,20 @@ define([
             return isPostFreetype() && $scope.selectedPostType.name === 'Scorecard';
         }
 
-
         //save the 'keep scoarers' if needed
         function saveScorers() {
-                var bp = angular.copy($scope.currentBlog.blog_preferences);
-                bp.last_scorecard = $scope.freetypesData;
-                return blogService.update($scope.currentBlog, {'blog_preferences': bp});
+                if ($scope.currentBlog.blog_preferences) {
+                    var bp = angular.copy($scope.currentBlog.blog_preferences);
+                    bp.last_scorecard = $scope.freetypesData;
+                    return blogService.update($scope.currentBlog, {'blog_preferences': bp});
+                } else {
+                    return blogService.get($scope.blog._id).then(function(currentBlog) {
+                        $scope.currentBlog = currentBlog;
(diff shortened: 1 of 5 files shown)


client/app/scripts/liveblog-edit/controllers/blog-edit.js
```diff
@@ -65,6 +65,25 @@ define([
...
+        }
+
+
+        //save the 'keep scoarers' if needed
+        function saveScorers() {
+            if (isPostScorecard()) {
+                var bp = angular.copy($scope.currentBlog.blog_preferences);
+                bp.last_scorecard = $scope.freetypesData;
+                blogService.update($scope.currentBlog, {'blog_preferences': bp}).then(function(nb) {
+                    //move along nothing to see here
```
m0g: Why do you have a promise returning nothing?

client/app/scripts/liveblog-edit/controllers/blog-edit.js
```diff
@@ -265,6 +297,8 @@ define([
             },
             publish: function() {
                 $scope.actionPending = true;
+                //save the keep scoreres setting( if needed)
+                saveScorers();
```
m0g: Here you should chain your promise to something. At least to verify that the save went well

client/app/scripts/liveblog-edit/controllers/blog-edit.js
```diff
@@ -174,6 +194,18 @@ define([
             selectPostType: function(postType) {
                 $scope.selectedPostType = postType;
                 $scope.toggleTypePostDialog();
+                if (isPostScorecard()) {
+                    blogService.get($scope.blog._id).then(function(currentBlog) {
```
m0g: What is that function doing?
What is the need of reloading the current blog?

vladnicoara: We keep the last scorecard (with the 'remember scores' status) in the blog preferences so we need to load that. It is also needed in case somebody else saved the last scorecard( and the remember scorers status of course)

client/app/scripts/liveblog-edit/controllers/blog-edit.js
vladnicoara: fixed

client/app/scripts/liveblog-edit/controllers/blog-edit.js
vladnicoara: fixed

PR #612, opened 2017-02-09
Lbsd 1262

darconny:
introduce a help box for freetypes and theme manager


nistormihai:
only failed
`1) editor embed: add a youtube iframe in the editor`
but that is due to the fork issue.

https://travis-ci.org/liveblog/liveblog/jobs/199932663

PR #618, opened 2017-02-13
LBSD-1272 limit freetype title length

vladnicoara:
In order to keep the design from breaking in the editor when a freetype is selected, I limited the length of the freetype title shown to 32 characters. The actual title in the CRUD is limited to 65 chars for the similar reasons
If this is not an option then we have to reevaluate the design.


m0g:
:+1:

mscam:
:+1:

PR #620, opened 2017-02-14
LBSD-1305 unpublish and save in contributions doesn't work

vladnicoara:


m0g:
If we merge this, the syndicated posts will be displayed in the ingest panel

m0g:
https://github.com/liveblog/liveblog/pull/672

PR #626, opened 2017-02-17
LBSD-1315: Remove advertisment posts from syndication

mscam:
Removed "Advertisment Remote" and "Advertisment Local" new post types from syndicated posts.


server/liveblog/syndication/syndication.py
```diff
@@ -25,6 +26,8 @@
     'updated': 'PUT',
     'deleted': 'DELETE'
 }
+SYNDICATION_EXCLUDED_ITEMS = ('Advertisment Local', 'Advertisment Remote')
```
djbrd-sourcefabric: Can we put these strings in constants somewhere?

server/liveblog/syndication/syndication.py
```diff
@@ -94,6 +97,12 @@ def send_syndication_post(self, post, action='created'):
             logger.warning('Not sending post "{}": syndicated content.'.format(post['_id']))
             return
 
+        items = extract_post_items_data(post)
```
djbrd-sourcefabric: I think that it would make sense to put exclusion logic in a separate method, like _post_for_syndication, because it will grow

server/liveblog/syndication/syndication.py
mscam: Yes. Project settings default is the best place

m0g:
:+1:

server/liveblog/syndication/syndication.py
```diff
@@ -94,6 +97,12 @@ def send_syndication_post(self, post, action='created'):
...
+        for item in items:
+            if item['group_type'] == 'freetype' and item['item_type'] in SYNDICATION_EXCLUDED_ITEMS:
+                logger.warning('Not sending post "{}": syndicated content contains excluded items.'.format(post['_id']))
```
djbrd-sourcefabric: This is an expected case, so logging with warning level is not appropriate, I think. Should be logger.debug

server/liveblog/syndication/syndication.py
djbrd-sourcefabric: If you find that you're checking for these post types in more than one place, you should probably put this logic in utils in a function which takes a post as a parameter, or simply have extract_post_data instead of extract_post_items_data.

mscam: That should be only in send_syndication_post resource method I guess

PR #629, opened 2017-02-20
Fix/remove slideshow tooltips

vladnicoara:


vladnicoara:
Fix merged from another PR

PR #640, opened 2017-02-23
Feature/freetype validation 2

nistormihai:


client/app/scripts/liveblog-edit/controllers/blog-edit.js
```diff
@@ -194,26 +198,28 @@ define([
...
-                $scope.selectedPostType = postType;
-                $scope.toggleTypePostDialog();
-                if (isPostScorecard()) {
-                    blogService.get($scope.blog._id).then(function(currentBlog) {
-                        $scope.currentBlog = currentBlog;
-                        if ($scope.currentBlog.blog_preferences.last_scorecard) {
-                            //load latest scorecard
-                            if ($scope.currentBlog.blog_preferences.last_scorecard.remember) {
-                                $scope.freetypesData = angular.copy($scope.currentBlog.blog_preferences.last_scorecard);
+                doOrAskBeforeIfEditorIsNotEmpty(cleanEditor).then(function() {
```
m0g: So if I understand well here, you execute the `doOrAskBeforeIfEditorIsNotEmpty` function, which calls `cleanEditor` as a callback and then returns a promise?

client/app/scripts/liveblog-edit/controllers/blog-edit.js
```diff
@@ -113,12 +113,16 @@ define([
         // ask in a modalbox if the user is sure to want to overwrite editor.
         // call the callback if user say yes or if editor is empty
         function doOrAskBeforeIfEditorIsNotEmpty(callback, msg) {
```
m0g: Why is `doOrAskBeforeIfEditorIsNotEmpty` expecting two parameters when you only send one?

client/app/scripts/liveblog-edit/directives.js
```diff
@@ -534,15 +542,74 @@ define([
...
         }])
-        .directive('freetypeLink', ['$compile', function($compile) {
+        .directive('freetypeText', ['$templateCache', function($templateCache) {
 
             return {
                 restrict: 'E',
-                template: '<input type="url" ng-model="link"/>',
-                controller: function() {
-                },
+                template: $templateCache.get('scripts/liveblog-edit/views/freetype-text.html'),
```
m0g: I still don't understand what that template cache is used for and why not using the standard approach to load a template.
Plus this will pose problem in the webpack branch: https://github.com/liveblog/liveblog/commit/492fc5ad993f3e368bcefeb43cae8123f5cff5e3

client/app/scripts/liveblog-edit/directives.js
```diff
@@ -534,15 +542,74 @@ define([
...
+                controller: ['$scope', function($scope) {
+                    $scope._id = _.uniqueId('text');
+                    if ($scope.initial !== undefined && $scope.text === '') {
+                        $scope.text = String($scope.initial);
+                    }
+                    if ($scope.number !== undefined) {
+                        $scope.$on('$destroy', $scope.$watch('text', function(value) {
```
m0g: There is a lot of repetition in here. Couldn't this be replaced by a loop?

client/app/scripts/liveblog-edit/directives.js
```diff
@@ -534,15 +542,74 @@ define([
...
+                    order: '@',
+                    initial: '@'
+                }
+            };
+        }])
+        .directive('freetypeLink', ['$templateCache', function($templateCache) {
+
+            return {
+                restrict: 'E',
+                template: $templateCache.get('scripts/liveblog-edit/views/freetype-link.html'),
```
m0g: Why using `$templaceCache` every time?

client/app/scripts/liveblog-edit/freetype.service.js
```diff
@@ -152,9 +154,10 @@ function(angular) {
                             return all;
                         }
                     });
-                    return '<li ng-repeat="' + iteratorName + ' in ' + SCOPE_FREETYPEDATA + '.' + vector + '">' +
-                            repeater + '<freetype-collection-remove index="$index" vector="' + SCOPE_FREETYPEDATA + '.' + vector + '"/>' +
-                            '</li><li><freetype-collection-add vector="' + SCOPE_FREETYPEDATA + '.' + vector + '"/></li>';
+                    collection = SCOPE_FREETYPEDATA + '.' + vector;
+                    return '<li ng-repeat="' + iteratorName + ' in ' + collection + '">' +
```
m0g: Surely there must a better way to do that. Why should template strings be injected in a service?

client/app/scripts/liveblog-edit/directives.js
nistormihai: not so much, code will be unclear.

client/app/scripts/liveblog-edit/directives.js
nistormihai: in the webpack branch @vladnicoara found a different approach.
this PR is not here to fix that, just to enhance with validation.

client/app/scripts/liveblog-edit/controllers/blog-edit.js
nistormihai: Yeah kinda weird, I've added the return `deferred`, and not sure what the callback was there.
I will check that and change it.

client/app/scripts/liveblog-edit/controllers/blog-edit.js
nistormihai: others calls are with two parameters.

client/app/scripts/liveblog-edit/freetype.service.js
nistormihai: transform a $dollar template into an angular template. 
the only way is to manipulate strings with regex.

client/app/scripts/liveblog-edit/directives.js
nistormihai: all files from `views/*` are compiled under `$templateCache` so why not use if it is there.

client/app/scripts/liveblog-edit/directives.js
m0g: Yes, but maybe we could work together and planning things ahead. Instead of having me sorting things out every time.

nistormihai: Agreed, but a refactor here could take more time/efficiency  then it should. And porting the code to the new structure was done and should be easy to be updated.

m0g: So then you can take of it

m0g:
:+1:

PR #643, opened 2017-02-24
Lp 49 - Restore theme manager access for solo users (SAAS)

m0g:
* Restore theme manager for `solo` users
* Hide open source alert message
* Hide github link in themes list
* Hide new theme button
* Hide download theme

### DIFF
--- a/client/app/scripts/liveblog-themes/module.js
+++ b/client/app/scripts/liveblog-themes/module.js
@@ -129,6 +129,7 @@
             themeBlogsModal: false,
             // this is used to when a blog is selected.
             selectedBlog: false,
+            isSolo: config.subscriptionLevel == 'solo',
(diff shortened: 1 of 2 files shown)


djbrd-sourcefabric:
:+1:

client/app/scripts/liveblog-themes/module.js
```diff
@@ -129,6 +129,7 @@
             themeBlogsModal: false,
             // this is used to when a blog is selected.
             selectedBlog: false,
+            isSolo: config.subscriptionLevel == 'solo',
```
vladnicoara: do you need type conversion here? if not, === is faster

m0g: Didn't see that, I'll correct it

PR #646, opened 2017-02-27
LBSD-1333 Error message when no blogs found

m0g:
Syndication attach new blog, there should be an error message if producer does not have any blog

### DIFF
--- a/client/app/scripts/liveblog-syndication/views/attach-syndicated-blogs-modal.html
+++ b/client/app/scripts/liveblog-syndication/views/attach-syndicated-blogs-modal.html
@@ -23,7 +23,10 @@
                 <i class="icon-refresh"></i>
             </button>
         </div>
-        <ul class="producers-list">
+        <div ng-if="producerBlogs._items.length == 0">
+            The selected producer hasn't made any blogs available for syndication yet.
+        </div>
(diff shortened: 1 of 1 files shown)


client/app/scripts/liveblog-syndication/views/attach-syndicated-blogs-modal.html
```diff
@@ -23,7 +23,10 @@ <h3 translate>Attach Syndicated blog</h3>
                 <i class="icon-refresh"></i>
             </button>
         </div>
-        <ul class="producers-list">
+        <div ng-if="producerBlogs._items.length == 0">
```
nistormihai: add `translate` attribute.

djbrd-sourcefabric:
:+1:

client/app/scripts/liveblog-syndication/views/attach-syndicated-blogs-modal.html
```diff
@@ -23,7 +23,10 @@ <h3 translate>Attach Syndicated blog</h3>
                 <i class="icon-refresh"></i>
             </button>
         </div>
-        <ul class="producers-list">
+        <div ng-if="producerBlogs._items.length == 0" translate>
```
nistormihai: I was wondering why this was passing jshint, then I realized that is in template.
Please use ===

PR #657, opened 2017-03-02
LBSD-1350 Bug when trying to type the URL manually

m0g:
> In authoring click to add an embed.
> Observe: as soon as the first character is placed in the embed field it is set and further typing is not possible.

My approach to fix that is to trigger the change function on paste action or when pressing enter/return. Therefore when somebody is typing something in the input field, the embed retrieval routine will only be triggered when pressing the return key


client/app/scripts/ng-sir-trevor-blocks.js
```diff
@@ -141,6 +141,7 @@ define([
                         input = fixSecureEmbed(input);
                         // if the input is an url, use embed services
                         if (isURI(input)) {
+                            console.log('is uri');
```
nistormihai: not needed

client/app/scripts/ng-sir-trevor-blocks.js
```diff
@@ -156,7 +157,15 @@ define([
                         } else {
                             that.loadData({html: input});
                         }
-                    }, 200));
+                    }
+                    this.$editor.on('paste', _.debounce(callServiceAndLoadData, 200));
+
+                    this.$editor.on('keydown', function(e) {
+                        if (e.keyCode == 13) {
```
nistormihai: should fail on jshint

m0g:
> but just to be sure we can have some sort of information regarding this ( placeholder or just an info text above when starting to type)

@nistormihai you mean to indicate the user that he needs to press `enter` to load the embed?

djbrd-sourcefabric:
:+1:

PR #662, opened 2017-03-05
LBSD-1370 Limit number of characters for the photographer field to ma…

vladnicoara:
…x. 300

### DIFF
--- a/client/app/scripts/ng-sir-trevor-blocks.js
+++ b/client/app/scripts/ng-sir-trevor-blocks.js
@@ -427,7 +427,6 @@ define([
                 icon_name: 'image',
                 descriptionPlaceholder: window.gettext('Add a description'),
                 authorPlaceholder: window.gettext('Add author / photographer'),
-                limitedText: window.gettext('Max. amount of 300 characters is reached'),
                 loadData: function(data) {
                     var file_url = (typeof(data.file) !== 'undefined') ? data.file.url : data.media._url, that = this;
                     this.$editor.html($('<img>', {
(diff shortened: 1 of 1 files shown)


client/app/scripts/ng-sir-trevor-blocks.js
```diff
@@ -438,12 +439,15 @@ define([
...
-                    
+
                     this.$editor.append($('<div>', {
                         name: 'credit',
                         class: 'st-image-block',
                         contenteditable: true,
-                        placeholder: that.authorPlaceholder
+                        placeholder: that.authorPlaceholder,
+                        oninput:
+                        'if(this.innerText.length > 300) { alert("' + that.limitedText + '"); this.innerText = this.innerText.substring(0, 300); this.style="border:1px solid red"}' +
```
m0g: What's going on here?

vladnicoara: Adding 'oninput' attribute for the image author. oninput works for paste as well as opposed to keyup or keydown. As for the location and the form of the code, I did not find a better way to work with this sir trevor element.

m0g: Yes, but do you really need to inject javascript code like that?

vladnicoara: That is the first thing I have tried of course. If I define a function there( when appending the input to the editor) or pass a defined one, it just gets called once on render and then nothing.
I found a way to add that functionality in proper way after it has been rendered

PR #669, opened 2017-03-07
LBSD-1402 Deactivate blog actions when archived

m0g:
Added the following when viewing an archived blog:

* Deactivate blog actions in the timeline (moved, delete, star...)
* Deactivate the Live link

### DIFF
--- a/client/app/scripts/liveblog-edit/directives.js
+++ b/client/app/scripts/liveblog-edit/directives.js
@@ -135,7 +135,8 @@ define([
                             return vm.pagesManager.setAuthors(users.map(function(user) {return user._id;})).then(function() {
                                 vm.isLoading = false;
                             });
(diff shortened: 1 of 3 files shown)


client/app/scripts/liveblog-edit/directives.js
```diff
@@ -135,7 +135,8 @@ define([
                             return vm.pagesManager.setAuthors(users.map(function(user) {return user._id;})).then(function() {
                                 vm.isLoading = false;
                             });
-                        }
+                        },
```
djbrd-sourcefabric: Any reason here not to use the isBlogOpened method in the BlogEditController?

m0g: I can't access it from the directive

PR #670, opened 2017-03-07
LBSD-1404 Deactivate create blog button when the blog is being created

m0g:


client/app/scripts/liveblog-bloglist/module.js
```diff
@@ -86,7 +86,11 @@
                 });
         };
 
+        $scope.creationInProcess = false;
```
djbrd-sourcefabric: Where is this set back to false? Is the module destroyed after creation?

djbrd-sourcefabric:
:+1:

PR #674, opened 2017-03-09
LBSD-1351 Sorted out embed instagram issue

m0g:
> take the embed URL for an Instagram post
> try publishing it
> Observe that while in preview the image appears within a relatively acceptable timeframe, once it gets published the image post does not appear in the right panel, unless another post is triggered from the left one.


djbrd-sourcefabric:
:+1:

PR #682, opened 2017-03-10
LBSD-1441 fix blog avatar workaround for adding too small image

vladnicoara:
fix for both blog settings and blog creation

### DIFF
--- a/client/app/scripts/liveblog-bloglist/module.js
+++ b/client/app/scripts/liveblog-bloglist/module.js
@@ -90,8 +90,8 @@
             var members = _.map($scope.blogMembers, function(obj) {
                 return {user: obj._id};
             });
-            //upload image if there is an image that was not invalidated by dimentions constrains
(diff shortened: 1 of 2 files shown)


client/app/scripts/liveblog-bloglist/module.js
```diff
@@ -90,7 +90,8 @@
             var members = _.map($scope.blogMembers, function(obj) {
                 return {user: obj._id};
             });
-            var promise = angular.equals({}, $scope.preview) ? $q.when() : $scope.upload($scope.preview);
+            //upload image if there is an image that was not invalidated by dimentions constrains
+            var promise = angular.equals({}, $scope.preview) || !$scope.preview.url ? $q.when() : $scope.upload($scope.preview);
```
m0g: I think your condition is a bit too long and complicated to be expressed as a ternary operator

PR #689, opened 2017-03-15
LBSD-1471 Override config variables without having to rebuild

m0g:
So here is my proposition for sorting out that production build issue.

You need to rename `config-override.sample.js` into `config-override.js` and then you can manually set each variable to suit your needs.


naspeh:
I think that it should be `superdesk` compatible way, liveblog is based on it. Don't like the idea to have two different ways to do the same thing on `superdesk` and `liveblog`

naspeh:
config should be `/opt/liveblog/client/dist/config.js` in the end:
```
root@lbpr-689:~# cat /opt/liveblog/client/dist/config.js
window.superdeskConfig = {
    //subscriptionLevel: 'solo', // SUBSCRIPTION_LEVEL
    syndication: true, // SYNDICATION
    //marketplace: true, // MARKETPLACE
    //maxContentLength: 8 * 1024 * 1024 // MAX_CONTENT_LENGTH
};
```

client/app/config.js
```diff
@@ -0,0 +1,10 @@
+// This file allows you to change the config value
+// without having to rebuild the application
+// To do so, simply rename this file to config-override.js
```
naspeh: BTW: comments is not actual anymore

PR #697, opened 2017-03-17
Eslint

m0g:
Updated to eslint by copying the configuration from superdesk


m0g:
Linting specs & tasks will be done later on https://dev.sourcefabric.org/browse/LBSD-1506

PR #699, opened 2017-03-18
LBSD-1509 Load templates with webpack

vladnicoara:
Before this, when templates where changed, the client server needed to be restarted or the copy:assets grunt task had to be run and the browser refreshed in order for the changes to be visible. Also no template cache was available for the the views. 

JIRA issue [LBSD-1509](https://dev.sourcefabric.org/browse/LBSD-1509)
### Objectives
- [x] Rebuild the bundle after template change
- [x] Automatically refresh the browser after template change
- [x] Use template cacheing mechanism for views
- [x] Remove the copying of views in dist folder for LB components

### Actions taken
- [x] Add extra webpack loader to cache templates
- [x] Switch off 'hot' building of the bundle
- [x] Import all views and activity templates 
- [x] Remove the copying of views/templates in dist folder for LB components from the copy:assets task

### DIFF
--- a/client/Gruntfile.js
+++ b/client/Gruntfile.js
@@ -46,7 +46,6 @@ module.exports = function (grunt) {
         'copy:assets',
         'copy:index',
         'copy:sirTrevor',
-        'ngtemplates:dev',
         'webpack-dev-server:start'
     ]);
 
@@ -66,7 +65,6 @@ module.exports = function (grunt) {
         'copy:assets',
         'copy:index',
         'copy:sirTrevor',
-        'ngtemplates:dev',
         'webpack:build'
     ]);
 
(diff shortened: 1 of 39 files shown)


vladnicoara:
Before this when templates where changed, the client server needed to be restarted or the copy:assets grunt task had to be run and the browser refreshed in order for the changes to be visible. Also no template cache was available for the the views. 

JIRA issue [LBSD-1509](https://dev.sourcefabric.org/browse/LBSD-1509)
### Objectives
- [x] Rebuild the bundle after template change
- [x] Automatically refresh the browser after template change
- [x] Use template cacheing mechanism for views
- [x] Remove the copying of views in dist folder for LB components

PR #701, opened 2017-03-19
Feat/ads vlad

vladnicoara:
main advertising view with menu main menu entry point and two tabs for adverts and collections, 
it also contains some webpack changes from another PR that enables much faster way of working with templates


client/app/scripts/liveblog-advertising/index.js
```diff
@@ -0,0 +1,2 @@
+import './styles/advertising.scss';
+import './module';
```
m0g: Hey @vladnicoara in order to setup good practices for the code base, when creating a new module,  could you follow the same architecture as the `liveblog-syndication` module?

The index file is where you import all the files including the CSS and define your module: https://github.com/liveblog/liveblog/tree/devel/client/app/scripts/liveblog-syndication. Then for each and every directive and controller, you need to create a sub file in the following fashion:

* Controller: https://github.com/liveblog/liveblog/blob/devel/client/app/scripts/liveblog-syndication/controllers/syndication.js
* Directive: https://github.com/liveblog/liveblog/blob/devel/client/app/scripts/liveblog-syndication/directives/consumer-list.js

PR #707, opened 2017-03-22
LBSD-1327 Integration with news agency feeds, drag & drop

m0g:
## Description

In the blog editor, when adding a image block, a user can now directly drag & drop and picture from any remote web page.

## What has been done:

* New rules for drag and drop from a remote URL
* Send the URL to backend
* Have the URL be processed on the backend and return an image resource
* Integrate this new image in a normal item
* Linting

https://dev.sourcefabric.org/browse/LBSD-1327


client/app/scripts/liveblog-edit/controllers/blog-edit.js
```diff
@@ -436,6 +436,30 @@ var BlogEditController = function (api, $q, $scope, blog, notify, gettext, sessi
                         success_callback({media: media_meta});
                     }, handleError);
                 });
+            },
+            gogoGadgetoRemoteImage: function(imgURL) {
```
nistormihai: :+1: nice inspector gadget

m0g: :-)

server/liveblog/items/items.py
```diff
@@ -131,3 +140,19 @@ def get(self, req, lookup):
...
+    url = data['image_url']
+    mimetype = data['mimetype']
+
+    item_data = dict()
+    item_data['type'] = 'picture'
+    item_data['media'] = FileStorage(stream=fetch_url(url), content_type=mimetype)
+    archive_service = get_resource_service('archive')
+    archive_id = archive_service.post([item_data])[0]
+    archive = archive_service.find_one(req=None, _id=archive_id)
+
```
nistormihai: where are `renditions`?
how is this different then superdesk `archive`?

djbrd-sourcefabric: It's not different to the superdesk archive. Just a different way of creating such an item. You send the url of the image to the server, and it is downloaded there. We can't add this to the archive code, as it's in superdesk. So, I created another endpoint to create an archive, reusing as much of the archive code as seems possible to me. The renditions are created by the archive code. If you can think of a better, cleaner way, let me know.

PR #712, opened 2017-03-24
LBSD-1472 Ask confirmation before removing consumer or producer

m0g:


client/app/scripts/liveblog-syndication/directives/consumer-list.js
```diff
@@ -16,37 +16,37 @@ export default function consumerList(api, notify) {
...
+                modal.confirm(gettext('Are you sure you want to refresh the api key?'))
+                    .then(() => api.save('consumers', consumer, {api_key: ''}))
+                    .then((result) => {
+                        notify.pop();
+                        notify.success(gettext('api key updated.'));
+                    })
+                    .catch((err) => {
+                        if (err) {
+                            notify.pop();
+                            notify.error(gettext('Fatal error!'));
```
nistormihai: why not `notify.error(err.message)` the one from server. 
If something is wrong they will know at lest what kind of error.

m0g: Is this message variable always present here

PR #716, opened 2017-03-28
Feat/advertisement e2e

vladnicoara:


client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -96,18 +108,128 @@ upload, $templateCache, freetypeService, modal) {
...
+            $scope.collections = data._items;
+            notify.info('Collections loaded');
+            $scope.collectionsLoading = false;
+        }, function(data) {
+            $scope.collectionsLoading = false;
+            notify.error(gettext('There was an error getting the adverts'));
+        })
+    }
+
+    function handleCollectionSaveSuccess() {
```
m0g: Same, shouldn't that be a promise?

client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -96,18 +108,128 @@ upload, $templateCache, freetypeService, modal) {
...
+        loadCollections();
+    }
+
+    function handleCollectionSaveError() {
+        notify.error(gettext('Something went wrong, please try again later!'), 5000)
+    }
+
+    $scope.openCollectionDialog = function(collection) {
+        // load all available adverts without showing any messages
+        loadAdverts(true);
```
m0g: Shouldn't you be waiting for the adverts to be loaded before, executing the next instructions?

client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -96,18 +108,128 @@ upload, $templateCache, freetypeService, modal) {
...
+        })
+        var newCollection = {
+            'name': $scope.collection.name,
+            'advertisements': advertisements
+        }
+        $scope.dialogCollectionLoading = true;
+
+        if ($scope.collection._id) {
+            // we are editing existing collection
+            api('collections').save($scope.collection, newCollection).then(function(data) {
```
m0g: Maybe you could simplify that code to not have to repeat the query twice

client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -96,18 +108,128 @@ upload, $templateCache, freetypeService, modal) {
...
+                handleCollectionSaveSuccess();
+            }, function(data) {
+                handleCollectionSaveError();
+            });
+        }
+    }
+
+    $scope.removeCollection = function (collection, $index) {
+        modal.confirm(gettext('Are you sure you want to remove this collection?')).then(function() {
+            api('collections').save(collection, {deleted: true}).then(function(data) {
```
m0g: Same, unnecessary nested promises in here

client/app/scripts/liveblog-advertising/views/main.html
```diff
@@ -53,46 +54,134 @@ <h6 class="pull-left">
...
+                    </div>
+                </div>
+            </div>
+        </div>
+
 	</div>
 </section>
 
+<!-- MODAL FOR ADDING EDITING ADVERTS -->
 <div sd-modal data-model="advertModalActive">
```
m0g: Could you break that html file into smaller directives?

I'm sure `advertModal` could go in its own directive and template

client/app/scripts/liveblog-advertising/views/main.html
```diff
@@ -53,46 +54,134 @@ <h6 class="pull-left">
...
-                <button type="button" class="btn btn-default pull-left" translate ng-click="cancelAdvertCreate()">Cancel</button>
-                <button type="button" class="btn btn-primary btn-info" translate ng-click="saveAdvert()"
-                ng-disabled="!advert.name || dialogAdvertLoading"
-                >Save</button>
-            </div>
+    </div>
+</div>
+
+<!-- MODAL FOR ADDING EDITIONG COLLECTIONS -->
+<div sd-modal data-model="collectionModalActive" class="advertising">
```
m0g: Same for this modal, it can go in a separate directive

client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -50,26 +59,33 @@ upload, $templateCache, freetypeService, modal) {
...
         loadAdverts();
     }
 
+    function handleAdvertSaveError() {
+        notify.error(gettext('Something went wrong, please try again later!'), 5000)
+    }
+
     $scope.removeAdvert = function (advert, $index) {
         modal.confirm(gettext('Are you sure you want to remove this advert?')).then(function() {
             api('advertisements').save(advert, {deleted: true}).then(function(data) {
```
vladnicoara: I tried it, but if done so, the modal functionality breaks( you have to push the OK button a second time to close the modal, even if the action takes place, or if using catch then the notification error is shown when pressing the Cancel button)

client/app/scripts/liveblog-advertising/views/main.html
vladnicoara: I am also sure it could go into a separate directive and template, the issues  here is that we are nesting the freetypes directives in here with their own scopes and it would complicate things. It can be done as another improvement task to move all modals from main views as most modules use this approach.

client/app/scripts/liveblog-advertising/views/main.html
vladnicoara: same as above " It can be done as another improvement task to move all modals from main views as most modules use this approach. "

client/app/scripts/liveblog-advertising/controllers/advertising.js
vladnicoara: same issue with the confirm modal form as described above

client/app/scripts/liveblog-advertising/controllers/advertising.js
vladnicoara: I fixed that, I wonder why it's here

client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -50,26 +59,33 @@ upload, $templateCache, freetypeService, modal) {
...
+            }
             $scope.advertsLoading = false;
         }, function(data) {
             $scope.advertsLoading = false;
             notify.error(gettext('There was an error getting the adverts'));
         })
     }
 
-    function handleSaveSuccess() {
+    function handleAdvertSaveSuccess() {
```
vladnicoara: I fixed that, I wonder why it's here

m0g: Actually  I was more thinking about something like:

```javascript
api('advertisements')
  .save($scope.advert, newAd)
  .then(handleAdvertSaveSuccess)
  .then(loadAdverts)
```

client/app/scripts/liveblog-advertising/controllers/advertising.js
m0g: Yes, but I don't see it being corrected, your promises are still being nested

client/app/scripts/liveblog-advertising/views/main.html
m0g: I mean you can pass whatever scope variable you need to the child directive. I don't see that being too difficult

client/app/scripts/liveblog-advertising/controllers/advertising.js
nistormihai: This is just a mater of how I like the code in a different way. I don't see a better practice.
the only good part is that creating another function isn't needed.
ex: 
`
api('advertisements')
  .save($scope.advert, newAd)
  .then(handleAdvertSaveSuccess)
`

client/app/scripts/liveblog-advertising/controllers/advertising.js
nistormihai: no it shouldn't be a promise.
is how it says a handler that is called in the success promise.

client/app/scripts/liveblog-advertising/controllers/advertising.js
nistormihai: yes, this should be called on promise returned by `loadAdverts`

client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -96,18 +108,128 @@ upload, $templateCache, freetypeService, modal) {
         if ($scope.advert._id) {
             // we are editing existing ad
             api('advertisements').save($scope.advert, newAd).then(function(data) {
-                handleSaveSuccess();
+                handleAdvertSaveSuccess();
```
nistormihai: why it should return a promise?

client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -96,18 +108,128 @@ upload, $templateCache, freetypeService, modal) {
...
+
+
+    //COLLECTIONS
+    function loadCollections() {
+        $scope.collectionsLoading = true;
+        api('collections').query({where: {deleted: false}}).then(function(data) {
+            $scope.collections = data._items;
+            notify.info('Collections loaded');
+            $scope.collectionsLoading = false;
+        }, function(data) {
```
nistormihai: just the request error handling.

client/app/scripts/liveblog-advertising/controllers/advertising.js
nistormihai: What query are you talking about?

client/app/scripts/liveblog-advertising/controllers/advertising.js
nistormihai: @m0g why nesting is a problem here? and what is the alternative?

client/app/scripts/liveblog-advertising/views/main.html
nistormihai: Split this into smaller directives isn't KISS and also isn't DRY at this point.

PR #727, opened 2017-03-30
Feature/advertisment output

nistormihai:


mscam:
There are some conflicting files, but I don't see any relevant issue on backend code.

PR #732, opened 2017-03-31
LP-62 Restrict requests to access blogs, when max number of team members has been reached already

m0g:
Limit number of access requests for a blog according to current team number & pending access requests

https://dev.sourcefabric.org/browse/LP-62


djbrd-sourcefabric:
:+1:

PR #736, opened 2017-04-03
Changed tooltip for deleting producer

djbrd-sourcefabric:


m0g:
@djbrd-sourcefabric you need to make your PRs from branches and not from forks, otherwise the tests won't pass

PR #737, opened 2017-04-03
Fix/help menu

vladnicoara:


m0g:
Actually, wouldn't it make more sense UI wise to put those link at the bottom of the sidebar?

vladnicoara:
@m0g about the position of the links, I just put them back were they were, the location was chosen by Darko and Karel I guess. We can of course request to move them if it makes more sense for them to be somewhere else.

PR #744, opened 2017-04-04
Remove travis encryption

m0g:
It's now possible to add environment variables directly via the admin interface, thus enabling the possibility of making PRs from forks.

https://travis-ci.org/liveblog/liveblog/settings


djbrd-sourcefabric:
:+1:

PR #749, opened 2017-04-05
Fix/scorers

vladnicoara:
fixed the multiple lines of scorers, fixed template rendering and fixed hiding elements with hide-render attribute


client/app/scripts/liveblog-edit/freetype.service.js
```diff
@@ -283,7 +283,7 @@ import './module';
                     wrapBefore = '',
                     wrapAfter = '';
                 obj2path(paths, data);
-                template = template.replace(/\<li([^>]*)\>(.*?)\<\/li\>/g, function(all, attr, repeater) {
+                template = template.replace(/<li([^>]*)>((.|\n)*?)<\/li>/g, function(all, attr, repeater) {
```
m0g: What is the regex doing?

nistormihai: just the same but with multiple line now.

PR #767, opened 2017-04-10
Fix/user avatar themes master

nistormihai:


server/liveblog/themes/themes_assets/angular/theme.json
```diff
@@ -2,7 +2,7 @@
     "label": "Angular Liveblog Toolkit",
     "name": "angular",
     "abstract": true,
-    "version": "1.4.38",
+    "version": "3.0.1",
```
m0g: 3.0.1 ?

m0g:
Guys, what version is this fix targeting?

server/liveblog/themes/themes_assets/angular/theme.json
nistormihai: yes after talks with Gideon, it was decided that is the best to increase the major version to match the liveblog repo.
Also an increase on minor version was omitted, in the future we need to release also the `angular`, `classic` and now `default` themes on liveblog release.

nistormihai:
You mean what liveblog version? it should be on the next release version.

PR #771, opened 2017-04-10
Lbsd 1408/9

m0g:
https://github.com/liveblog/liveblog/pull/710


mscam:
we should add a test case on behave tests.

mscam:
This has been merged, but behave test is missing.

PR #775, opened 2017-04-11
LBSD-1557 Fetch blogs on membership's request approval

m0g:
https://dev.sourcefabric.org/browse/LBSD-1557


client/app/scripts/liveblog-bloglist/module.js
```diff
@@ -1,612 +0,0 @@
-import mainTemplate from 'scripts/liveblog-bloglist/views/main.html';
```
nistormihai: where is this gone?

m0g: Here https://github.com/liveblog/liveblog/blob/78751ece9a8d7eaa1b8b5d366d8f9c3b5aaf50de/client/app/scripts/liveblog-bloglist/index.js#L4

PR #777, opened 2017-04-12
Revert superdesk-core to version 1.6

mscam:
Updated requirements.txt to use superdesk-core version 1.6 instead of 1.7.


djbrd-sourcefabric:
:+1:

PR #780, opened 2017-04-12
CRUD for channel outputs

vladnicoara:


client/app/scripts/liveblog-edit/controllers/blog-settings.js
```diff
@@ -11,6 +11,8 @@
 import angular from 'angular';
 import _ from 'lodash';
 
+import outputModalTpl from 'scripts/liveblog-edit/views/output-modal.html';
```
m0g: So, now that we have webpack and that it very easy to require new file, can you please put the output modal into a separate directive. I really don't want to have to solve merge conflicts on a 500+ lines JS file.

PR #786, opened 2017-04-13
Called patch instead of update on posts_service in consumer webhook s…

djbrd-sourcefabric:
…o that notifications are triggered in on_updated


m0g:
I added the front end checks as well for non auto published & auto published blogs

djbrd-sourcefabric:
_is_post_for_syndication is called in syndication.send_syndication_post before sending a post to a consumer. The method checks whether the post has come from a producer. If so, it won't be sent on. So chaining should not be an issue.

PR #792, opened 2017-04-19
Feature/advertisment amazon

nistormihai:


server/liveblog/blogs/commands.py
```diff
@@ -8,15 +8,20 @@ class PublishBlogsCommand(superdesk.Command):
...
 
         # Retrieves all opened blogs.
         blogs_service = get_resource_service('blogs')
         blogs = blogs_service.get(req=None, lookup=dict(blog_status='open'))
+        outputs = blogs_service.get(req=None, lookup=dict(deleted=False))
         # Republish on s3.
         print('\n* Republishing blogs:\n')
         for blog in blogs:
-            url = publish_blog_embed_on_s3(blog_id=str(blog['_id']), safe=False)
+            for output in outputs:
```
djbrd-sourcefabric: No need to do a loop within a loop here. Just loop through the blogs and publish the default output, then loop through the outputs, in each of which you have a reference to a blog. You could even search for outputs with relations to a blog which hasn't been deleted. I think the syntax would be blog.blog_status

server/liveblog/blogs/embeds.py
```diff
@@ -85,13 +85,21 @@ def render_bloglist_embed(api_host=None, assets_root=None):
     return render_template('blog-list-embed.html', **scope)
 
 
-@embed_blueprint.route('/embed/<blog_id>')
-def embed(blog_id, api_host=None, theme=None):
+@embed_blueprint.route('/embed/<blog_id>', defaults={'theme': None, 'output': None})
+@embed_blueprint.route('/embed/<blog_id>/<theme>', defaults={'output': None})
```
djbrd-sourcefabric: Why support the definition of a theme in the url? Is this just for debugging?

server/liveblog/blogs/schema.py
```diff
@@ -41,6 +41,9 @@
     'public_url': {
         'type': 'string'
     },
+    'public_urls': {
```
djbrd-sourcefabric: Is there any need to store all the urls in the schema? This seems like a recipe for bugs as you are effectively storing the data in two different places - once in the outputs and once here.

server/liveblog/blogs/commands.py
nistormihai: there isn't a `default` output just the blog embed itself without any `output`.
I think this was is a bit clear what is happening, the number of publishing to s3 will be the same.

server/liveblog/blogs/embeds.py
nistormihai: this is development use only, but the s3 url link is constructed the same.
Theme is added due to multiple theme support and also forward planing for theme support on output.

server/liveblog/blogs/schema.py
nistormihai: the `s3` urls needs to be stored at some level.
I choose to stored them in `blog` level instead of the `output` level, because it is also near to the `multiple themes` implementation.

server/liveblog/blogs/tasks.py
```diff
@@ -21,11 +21,12 @@
 logger = logging.getLogger('superdesk')
 
 
-def publish_embed(blog_id, api_host=None, theme=None):
+def publish_embed(blog_id, theme=None, output=None, api_host=None):
```
djbrd-sourcefabric: I think it might make sense to create separate methods instead trying to handle 2 or 3 cases in one method. So you could create a publish_output_embed method.

server/liveblog/blogs/tasks.py
```diff
@@ -36,37 +37,92 @@ def publish_embed(blog_id, api_host=None, theme=None):
     return superdesk.upload.url_for_media(file_id)
 
 
-def delete_embed(blog_id):
+def delete_embed(blog_id, theme=None, output=None):
```
djbrd-sourcefabric: Again, doesn't seem to be much of an advantage to handle different cases all in one method.

server/liveblog/blogs/commands.py
nistormihai: `blog.blog_status` is used to know if the blog is archived.
The delete of a blog removes the blog and also the s3 embed.

also having the loop here for outputs can make that easier.

server/liveblog/blogs/tasks.py
nistormihai: this is handling the `multiple themes` cases a swell which is directly  linked to the publish url on the s3.

server/liveblog/blogs/tasks.py
nistormihai: it is not about the advantages but instead of the workflow.

djbrd-sourcefabric: Yeah, wouldn't it be clearer and easier to manage with more than one method? And it's not clear in the declaration that you can use the method to delete either the output or the theme, but not both. I mean, you really have to read the code to understand what it does. With separate methods, the behaviour would be clearer.

server/liveblog/blogs/tasks.py
```diff
@@ -36,37 +37,92 @@ def publish_embed(blog_id, api_host=None, theme=None):
...
+        # get the output `_id`
+        output_id = str(output.get('_id'))
+
+        # compile a theme if there is an `output`.
+        if output.get('theme'):
+            theme = output.get('theme')
+        else:
+            theme = blog['blog_preferences'].get('theme')
+
     if blog['blog_preferences'].get('theme', False):
```
djbrd-sourcefabric: I realise that this is not new code, but is it expected that this variable will ever not be set? I think not, in which case this code fails silently.

server/liveblog/blogs/tasks.py
```diff
@@ -36,37 +37,92 @@ def publish_embed(blog_id, api_host=None, theme=None):
...
+                                       theme,
+                                       output,
+                                       api_host='//{}/'.format(app.config['SERVER_NAME']))
+            public_urls = blog.get('public_urls', {'output': {}, 'theme': {}})
+
+            if output_id:
+                public_urls['output'][output_id] = public_url
+                get_resource_service('blogs').system_update(blog_id, {'public_urls': public_urls}, blog)
+                push_notification('blog', published=1, blog_id=blog_id, public_urls=public_urls)
+            elif theme:
```
djbrd-sourcefabric: Again, not clear in the method signature that you define either the theme or the output but not both.

server/liveblog/client_modules/__init__.py
```diff
@@ -13,6 +15,18 @@ def init_app(app):
     service = ClientPostsService(endpoint_name, backend=superdesk.get_backend())
     ClientPostsResource(endpoint_name, app=app, service=service)
 
+    endpoint_name = 'client_collections'
```
djbrd-sourcefabric: Should this not be 'client_advertisement_collections'? Just 'client_collections' is very generic and doesn't really describe what it is being represented.

server/liveblog/blogs/tasks.py
nistormihai: Agreed will refactor.

server/liveblog/themes/themes.py
```diff
@@ -271,17 +271,24 @@ def publish_related_blogs(self, theme):
         # for t in self.get_children(theme['name']) + [theme['name']]:
         #     terms.append({'term': {'blog_preferences.theme': t}})
         blogs = get_resource_service('blogs').get(req=None, lookup={})
```
djbrd-sourcefabric: Actually, you could get blogs which use this theme using the following lookup: {'blog_preferences.theme': theme['name']}

server/liveblog/themes/themes.py
```diff
@@ -271,17 +271,24 @@ def publish_related_blogs(self, theme):
...
+        outputs = get_resource_service('outputs').get(req=None, lookup={})
         # get all the children for the theme that we modify the settings for
         theme_children = self.get_children(theme.get('name'))
         for blog in blogs:
             blog_pref = blog.get('blog_preferences')
-            if blog_pref['theme'] == theme['name']:
+            if blog_pref.get('theme') == theme['name']:
```
djbrd-sourcefabric: When dealing with each individual blog which uses the theme which has been changed, you should then get outputs for this blog which use the same theme (i.e. which have a null theme)

server/liveblog/themes/themes.py
```diff
@@ -271,17 +271,24 @@ def publish_related_blogs(self, theme):
...
+                for output in outputs:
+                    if output.get('blog') == blog.get('_id'):
+                        publish_blog_embed_on_s3.delay(str(blog['_id']), output=output)
                 publish_blog_embed_on_s3.delay(str(blog['_id']))
             if theme_children:
                 # if a blog has associated the theme that is a  child of the one
                 # for which we modify the settings, we redeploy the blog on s3
                 for child in theme_children:
-                    if blog_pref['theme'] == child:
+                    if blog_pref.get('theme') == child:
```
djbrd-sourcefabric: You could call this function recursively with the theme's children. Or, probably better, create a local sub method which is called first with the given theme, and then with each of that theme's children. Then you don't have to duplicate the code in the if theme_children block.

server/liveblog/blogs/tasks.py
nistormihai: `blog_preferences` should be set with the theme from the creation of a blog.
This code maybe is for legacy.

server/liveblog/blogs/tasks.py
```diff
@@ -36,37 +37,92 @@ def publish_embed(blog_id, api_host=None, theme=None):
...
+    output_id = None
+    if output:
+        # get the output `_id`
+        output_id = str(output.get('_id'))
+
+        # compile a theme if there is an `output`.
+        if output.get('theme'):
+            theme = output.get('theme')
+        else:
+            theme = blog['blog_preferences'].get('theme')
```
nistormihai: necessary, as if a output is defined the theme of the output should be used if any otherwise a theme must be used and that theme is the blog theme.
I don't see that the theme management is handled bellow

server/liveblog/client_modules/__init__.py
nistormihai: agreed will change.

server/liveblog/themes/themes.py
nistormihai: children are involved as seen above.

server/liveblog/themes/themes.py
```diff
@@ -271,17 +271,24 @@ def publish_related_blogs(self, theme):
         # for t in self.get_children(theme['name']) + [theme['name']]:
         #     terms.append({'term': {'blog_preferences.theme': t}})
         blogs = get_resource_service('blogs').get(req=None, lookup={})
+        outputs = get_resource_service('outputs').get(req=None, lookup={})
```
nistormihai: an output can be defined without a theme.

nistormihai: also children are involved.

server/liveblog/themes/themes.py
nistormihai: as the code below shows, we get all the outputs based by the blog._id.

server/liveblog/themes/themes.py
nistormihai: it is better with kiss.
`get_children` gathers the children is recursively.

PR #795, opened 2017-04-20
Fixed syndication errors

djbrd-sourcefabric:


server/liveblog/syndication/consumer.py
```diff
@@ -79,7 +79,7 @@ def _get_consumer(self, consumer):
             consumer = self.find_one(_id=consumer, req=None)
         return consumer
 
-    def send_webhook_request(self, consumer_id, consumer_blog_token=None, method='GET', data=None, json_loads=True,
+    def _send_webhook_request(self, consumer_id, consumer_blog_token=None, method='GET', data=None, json_loads=True,
```
nistormihai: if this method was renamed, where is now called?

PR #813, opened 2017-04-28
fix: post type dropdown position

darconny:


client/app/scripts/liveblog-edit/views/main.html
```diff
@@ -217,7 +217,7 @@ <h3 class="page-nav-title">{{ blog.title }}</h3>
                             </div>
                         </div>
                     </div>
-                    <div class="freetype-selector" ng-click="$event.stopPropagation();" ng-if="selectPostTypeDialog">
+                    <div class="freetype-selector" ng-click="$event.stopPropagation();" ng-if="selectPostTypeDialog" ng-class="{'default': selectedPostType === 'Default'}">
```
m0g: Is it normal `'default': selectedPostType === 'Default'` is written with a capital D?

darconny: That's how it was in another tag, I suppose it was set in js like that.

m0g: Yeah you're right. That's why I approved the PR

darconny: OK I have to move the origin, for tests to pass.

PR #817, opened 2017-04-28
fix: edit freetype template.

nistormihai:


nistormihai:
LBSD-1604

mscam:
Some tests are not passing, but I am not sure if those errors are caused by your changes.

PR #818, opened 2017-05-01
LBSD-1609 Outputs CRUD e2e

vladnicoara:


client/spec/blog_settings_spec.js
```diff
@@ -184,4 +184,44 @@ describe('Blog settings', function() {
             blogs.expectCount(3);
         });
     });
+
+    fit('should do CRUD operations on output channels', function() {
```
nistormihai: should be `it`

PR #823, opened 2017-05-02
Fix/label before plus

nistormihai:


server/liveblog/themes/themes_assets/classic/main.js
```diff
@@ -33,8 +33,33 @@
...
+                        config.output = output;
+                        applyOutputStyle();
+                    }
+                })
+            }
+        }
+
+        function fixBackgroundImage(style) {
+            if (style['background-image']) {
+                style['background-image'] = 'url(' + style['background-image'] + ')';
```
m0g: Actually not mandatory, but with es2015, you can do:

```javascript
style['background-image'] = `url(${style['background-image']}')`;
```

nistormihai: this need to be remove actually, It is not in the release.
also themes are not ECMAScript 6

PR #826, opened 2017-05-03
Feature/advertisment

nistormihai:


client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -0,0 +1,241 @@
...
+        $scope.advertModalActive = true;
+    }
+    $scope.cancelAdvertCreate = function() {
+        $scope.freetypeControl.reset();
+        $scope.advertModalActive = false;
+    }
+
+    function loadAdverts(silent) {
+        silent = silent || false;
+        !silent ? $scope.advertsLoading = true : $scope.advertsLoading = false;
```
m0g: A suggestion:

```javascript
$scope.advertsLoading = !silent
```

client/app/scripts/liveblog-edit/controllers/blog-settings.js
```diff
@@ -48,12 +53,21 @@ function BlogSettingsController(
...
-    superdesk
+    superdesk,
+    urls,
+    $rootScope
 ) {
 
     // set view's model
     var vm = this;
 
+    var notif_listener = $rootScope.$on('blog', (e, data) => {
```
nistormihai: event fired from web sockets are send via $rootScope.

client/app/scripts/liveblog-advertising/ads-util.service.js
```diff
@@ -0,0 +1,23 @@
...
+		uniqueNameInItems: uniqueNameInItems
+	}
+
+	function uniqueNameInItems (item, existingItems) {
+        var invalid = false;
+        if (!item || !item.name) {
+            invalid = gettext('required');
+        }
+        if (!invalid) {
+            angular.forEach(existingItems, function(existingItem) {
```
djbrd-sourcefabric: From an answer on stackoverflow, looks like you do this quite a lot more elegantly with es6. See the 4th answer here: http://stackoverflow.com/questions/12462318/find-a-value-in-an-array-of-objects-in-javascript

client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -0,0 +1,241 @@
...
+        collection = collection || false;
+        // load all available adverts without showing any messages
+        loadAdverts(true).then(function() {
+            if (collection) {
+                // editing collection
+                $scope.collection = angular.copy(collection);
+                $scope.collection.checkAdverts = {};
+                //console.log('$scope.collection ', $scope.collection);
+                angular.forEach($scope.adverts, function(advert) {
+                    if ($scope.collectionHasAdvert($scope.collection, advert)) {
```
djbrd-sourcefabric: Can you not just write this block in one line? You don't need an if/else to set a boolean

client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    $scope.openCollectionDialog = function(collection) {
+        collection = collection || false;
+        // load all available adverts without showing any messages
+        loadAdverts(true).then(function() {
+            if (collection) {
+                // editing collection
+                $scope.collection = angular.copy(collection);
+                $scope.collection.checkAdverts = {};
```
djbrd-sourcefabric: Should be an array, not an object

client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -0,0 +1,241 @@
...
+            }
+            $scope.collectionModalActive = true;
+        });
+    }
+
+    $scope.saveCollection = function() {
+
+        //create the saveable advertisement array for the collection
+        var advertisements = [];
+        angular.forEach($scope.collection.checkAdverts, function(checked, ad_id) {
```
djbrd-sourcefabric: So you just use this array of booleans to create a new array? Why not just create the new array without creating an array of booleans?

client/app/scripts/liveblog-advertising/controllers/advertising.js
djbrd-sourcefabric: Also, to create such an array, why not create an array with all advert ids as keys, and all values set to false, then loop through the adverts in the collection, and set those to true? Then you would avoid a nested loop.

client/app/scripts/liveblog-advertising/controllers/advertising.js
```diff
@@ -0,0 +1,241 @@
...
+        return hasAdvert;
+    }
+
+    $scope.notValidName = adsUtilSevice.uniqueNameInItems;
+
+    $scope.freetypeValid = function() {
+        var valid = true;
+        angular.forEach($scope.validation, function(validStatus) {
+            if (!validStatus) {
+                valid = false;
```
djbrd-sourcefabric: You could presumably return here, and not loop through the entire array.

server/liveblog/themes/themes.py
```diff
@@ -275,17 +275,24 @@ def publish_related_blogs(self, theme):
         # for t in self.get_children(theme['name']) + [theme['name']]:
         #     terms.append({'term': {'blog_preferences.theme': t}})
         blogs = get_resource_service('blogs').get(req=None, lookup={})
+        outputs = get_resource_service('outputs').get(req=None, lookup={})
```
djbrd-sourcefabric: Is there no reference to an blog's outputs in the blog's resource?

server/liveblog/themes/themes.py
```diff
@@ -275,17 +275,24 @@ def publish_related_blogs(self, theme):
...
         # get all the children for the theme that we modify the settings for
         theme_children = self.get_children(theme.get('name'))
         for blog in blogs:
             blog_pref = blog.get('blog_preferences')
-            if blog_pref['theme'] == theme['name']:
+            if blog_pref.get('theme') == theme['name']:
+                for output in outputs:
```
djbrd-sourcefabric: Again, you could easily avoid this nested loop, I think

server/liveblog/themes/themes.py
```diff
@@ -275,17 +275,24 @@ def publish_related_blogs(self, theme):
...
+                    if output.get('blog') == blog.get('_id'):
+                        publish_blog_embed_on_s3.delay(str(blog['_id']), output=output)
                 publish_blog_embed_on_s3.delay(str(blog['_id']))
             if theme_children:
                 # if a blog has associated the theme that is a  child of the one
                 # for which we modify the settings, we redeploy the blog on s3
                 for child in theme_children:
-                    if blog_pref['theme'] == child:
+                    if blog_pref.get('theme') == child:
+                        for output in outputs:
```
djbrd-sourcefabric: Another nested loop

server/liveblog/posts/posts.py
```diff
@@ -26,7 +26,8 @@ def private_draft_filter():
     """
     private_filter = {'should': [{'term': {'post_status': 'open'}},
                                  {'term': {'post_status': 'submitted'}},
-                                 {'term': {'post_status': 'comment'}}]}
+                                 {'term': {'post_status': 'comment'}},
```
djbrd-sourcefabric: Very strange to have 'comment' as a possible status for a post, no? Is 'comment' a status like 'open' or 'submitted'? Seems more like a type.

client/app/scripts/liveblog-advertising/controllers/advertising.js
nistormihai: this is how angular works, this array is use for check boxes status.

client/app/scripts/liveblog-advertising/controllers/advertising.js
nistormihai: the same a above.

client/app/scripts/liveblog-advertising/controllers/advertising.js
nistormihai: no, an object with _ids for keys is the way to go for working with check boxes.

client/app/scripts/liveblog-advertising/controllers/advertising.js
nistormihai: true to Boolean assignation I will fix that.

server/liveblog/client_modules/client_modules.py
```diff
@@ -86,6 +89,57 @@ class ClientPostsService(PostsService):
     pass
 
 
+class ClientCollectionsResource(CollectionsResource):
+    datasource = {
+        'source': 'collections',
+        'default_sort': [('name', 1)]
```
djbrd-sourcefabric: Is the reason for this resource just to provide a shortcut for a request with a certain kind of sort? In which case, it's a lot of code to save having to add a parameter on the front end.

server/liveblog/blogs/utils.py
```diff
@@ -10,8 +10,10 @@ def check_media_storage():
         raise MediaStorageUnsupportedForBlogPublishing()
 
 
-def get_blog_path(blog_id):
-    return 'blogs/%s/index.html' % (blog_id)
+def get_blog_path(blog_id, theme=None, output_id=None):
+    return 'blogs/{}/{}{}index.html'.format(blog_id,
+                                            '{}/'.format(theme) if theme else '',
```
djbrd-sourcefabric: Should it be possible to create a blog with a different theme without creating a separate output?

server/liveblog/blogs/embeds.py
```diff
@@ -85,13 +85,21 @@ def render_bloglist_embed(api_host=None, assets_root=None):
     return render_template('blog-list-embed.html', **scope)
 
 
-@embed_blueprint.route('/embed/<blog_id>')
-def embed(blog_id, api_host=None, theme=None):
+@embed_blueprint.route('/embed/<blog_id>', defaults={'theme': None, 'output': None})
+@embed_blueprint.route('/embed/<blog_id>/<theme>', defaults={'output': None})
+@embed_blueprint.route('/embed/<blog_id>/<theme>/<output>')
+def embed(blog_id, theme=None, output=None, api_host=None):
```
djbrd-sourcefabric: Is it possible to override a theme without creating an output? If so, why?

server/liveblog/blogs/commands.py
```diff
@@ -8,15 +8,20 @@ class PublishBlogsCommand(superdesk.Command):
...
         blogs = blogs_service.get(req=None, lookup=dict(blog_status='open'))
+        outputs = blogs_service.get(req=None, lookup=dict(deleted=False))
         # Republish on s3.
         print('\n* Republishing blogs:\n')
         for blog in blogs:
-            url = publish_blog_embed_on_s3(blog_id=str(blog['_id']), safe=False)
+            for output in outputs:
+                if output['blog'] == blog['_id']:
+                    url = publish_blog_embeds_on_s3(blog_id=str(blog['_id']), output=output, safe=False)
+                    print('  - Blog "%s" output "%s" republished: %s' % (blog['title'], output['name'], url))
```
djbrd-sourcefabric: I guess, remove the debug code

server/liveblog/blogs/tasks.py
```diff
@@ -36,37 +37,92 @@ def publish_embed(blog_id, api_host=None, theme=None):
...
+        public_urls = {}
+
     # Remove existing file.
-    app.media.delete(file_path)
+    app.media.delete(public_url)
+    get_resource_service('blogs').system_update(blog_id, {'public_urls': public_urls}, blog)
 
 
-def _publish_blog_embed_on_s3(blog_id, safe=True):
+def _publish_blog_embed_on_s3(blog_id, theme=None, output=None, safe=True):
```
djbrd-sourcefabric: Should 'safe' not be false by default? I think you should have to explicitly suppress or handle the exception. Otherwise you risk silent failure.

server/liveblog/advertisements/collections.py
```diff
@@ -0,0 +1,53 @@
...
+    schema = {
+        'name': {
+            'type': 'string',
+            'unique': True
+        },
+        'advertisements': {
+            'type': 'list',
+            'schema': {
+                'type': 'dict',
+                'schema': {
```
djbrd-sourcefabric: You seriously want a 'schema' key in a dictionary called 'schema'

server/liveblog/advertisements/advertisements.py
```diff
@@ -0,0 +1,55 @@
...
+
+class AdvertisementsResource(Resource):
+    schema = {
+        'name': {
+            'type': 'string',
+            'unique': True
+        },
+        'type': {
+            'type': 'string',
+            'allowed': ['Advertisement Local', 'Advertisement Remote', 'Advertisement Adsense'],
```
djbrd-sourcefabric: Why prefix each types with 'Advertisement'?

client/app/scripts/liveblog-advertising/ads-util.service.js
nistormihai: I will make the change, but a review should be on how elegant it is the code.

client/app/scripts/liveblog-advertising/controllers/advertising.js
nistormihai: not how js works :(, but a reducer or a find can be used here.

server/liveblog/themes/themes.py
nistormihai: no, it shouldn't be the same as there isn't a reference to blog's pots in the blog's resource.

server/liveblog/posts/posts.py
nistormihai: this comment has nothing with the PR.

server/liveblog/blogs/utils.py
nistormihai: yes, this is how was thought at the beginning when multiple themes where in place.
now we can have another behavior but this isn't breaking that.

server/liveblog/blogs/commands.py
nistormihai: it isn't a debug code, it is a info message for the command.

server/liveblog/advertisements/collections.py
nistormihai: that it is not how this works.

server/liveblog/advertisements/advertisements.py
nistormihai: it should match the type in the item type, for rendering purposes.

server/liveblog/themes/themes.py
nistormihai: As in the previous PR comments showed, this isn't the case.

PR #834, opened 2017-05-08
Added missing label for create blog btn

m0g:


darconny:
Please check devel instance, now we have same sd-create-btn (with title shown on hover) on all views (themes, users, blog list). We do not need this additional label.

PR #861, opened 2017-05-24
LBSD-1617 Publish button not visible after canceling large image upload

vladnicoara:
"Add content" button is invisible after halting the upload process of a
larger image implying that termination is not working


client/app/scripts/liveblog-edit/controllers/blog-edit.js
```diff
@@ -457,6 +457,9 @@ export default function BlogEditController(
                     $scope.$digest();
                 }
             },
+            notPending: function() {
```
m0g: What about `setPending: function(value) {` ?

PR #862, opened 2017-05-24
Removing date from syndication UI

m0g:


djbrd-sourcefabric:
:+1: hard to argue with a lot of deletion

PR #870, opened 2017-05-29
LBSD-1704: Handle uploaded themes in a different folder

mscam:
- Handle uploaded themes in a different folder.
- Remove uploaded files from git subtrees themes directory.
- Remove uploaded files when s3 storage is enabled.
- Refactor theme resource and embed renderer.
- Use theme upload directory when s3 storage is disabled.
- Add setting to specify custom theme upload directory.
- Serve uploaded files when s3 storage is disabled.
- Store theme embed template on db.
- Pre-compile jinja2 templates for SEO themes.
- Add setting to specify custom jinja2 pre-complied templates directory.
- Update .gitignore.


server/liveblog/themes/themes.py
```diff
@@ -157,148 +170,286 @@ def get_options(self, theme, options=None):
...
-                    files.append(os.path.join(root, filename))
-            yield json.loads(open(file).read()), files
+        """
+        Get local instance theme packages.
+        It includes UPLOAD_THEMES_DIRECTORY if s3 storage is disabled.
+
+        :return:
+        """
+        theme_dirs = [LOCAL_THEMES_DIRECTORY]
+        if not self.is_s3_storage_enabled:
```
nistormihai: so this means that if s3 is enabled we don't keep the theme files in the file storage?

server/liveblog/themes/themes.py
```diff
@@ -157,148 +170,286 @@ def get_options(self, theme, options=None):
...
+            # Set the content type
+            mime = magic.Magic(mime=True)
+            content_type = mime.from_file(name)
+            if content_type == 'text/plain' and name.endswith(tuple(CONTENT_TYPES.keys())):
+                content_type = CONTENT_TYPES[os.path.splitext(name)[1]]
+
+            final_file_name = os.path.relpath(name, CURRENT_DIRECTORY)
+
+            # TODO: Add version parameter to media_id() after merging related core-changes in amazon_media_storage
+            # and desk_media storage.
```
nistormihai: why is this version variable isn't good?

nistormihai: also don't forget the
 ```
                    # don't use version for the `parent-iframe.js` script
                    # requering the version for this will be done from the client.
                    if name.endswith('parent-iframe.js'):
                        version = False
```

server/liveblog/themes/themes.py
mscam: Yes this is made to prevent to have uploaded files in the same directory where default themes are stored (as git subtree) and to prevent to keep the theme files in the local fs. This is actually breaking zip theme download

server/liveblog/themes/themes.py
mscam: @nistormihai "version" related code was commented before I introduced SEO theme features, we should check why it was commented before

mscam:
There is still one issue to manage the download of s3-only uploaded theme

mscam:
This cannot actually be merged as it is breaking classic theme and some admin features.

PR #874, opened 2017-05-30
Various notification fixes for syndication

m0g:
* Should not update the notification count in ingest panel when publishing an ingested post
* Should reset the notification count in ingest panel when leaving the view
* Fix broken tooltip notification count (ingest logo)


djbrd-sourcefabric:
:+1:

PR #914, opened 2017-07-17
Remove uploaded theme files when s3 is enabled (LBSD-1704)

mscam:
- Handle uploaded themes in a different folder.
- Remove uploaded files from git subtrees themes directory.
- Remove uploaded files when s3 storage is enabled.
- Refactor theme resource and embed renderer.
- Use theme upload directory when s3 storage is disabled.
- Add setting to specify custom theme upload directory.
- Serve uploaded files when s3 storage is disabled.
- Store theme embed template on db.
- Pre-compile jinja2 templates for SEO themes.
- Add setting to specify custom jinja2 pre-complied templates directory.
- Update .gitignore.


mscam:
This is not currently allowing to update an existing theme. We should prevent overwrite of default, amp and classic only.

PR #921, opened 2017-08-09
quote fix for the editor and preview

aleksandarjelicic:


nistormihai:
please use liveblog/liveblog branches, when PR from fork it will fail for youtube embed.

PR #931, opened 2017-09-06
Finished permalink and bug fixes

Aca-jov:


nistormihai:
need to be ported into 
https://github.com/liveblog/liveblog-default-theme

PR #941, opened 2017-09-25
Update superdesk-core with fix for [SDESK-1957]

naspeh:
Ref: https://dev.sourcefabric.org/browse/SDESK-1957


nistormihai:
only failed e2e for youtube adding.

PR #942, opened 2017-09-25
Update superdesk-core with fix for [SDESK-1957]

naspeh:
Ref: https://dev.sourcefabric.org/browse/SDESK-1957


naspeh:
So I'm closing this, because `devel` branch needs some care first, it also need other parts from `release/3.3.0` as well

PR #978, opened 2017-10-30
fix: mappings and publisher.

nistormihai:
publisher clean up.

### DIFF
--- a/server/liveblog/advertisements/advertisements.py
+++ b/server/liveblog/advertisements/advertisements.py
@@ -25,6 +25,10 @@ class AdvertisementsResource(Resource):
         },
         'meta': {
             'type': 'dict',
+            'mapping': {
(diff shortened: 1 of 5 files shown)


server/liveblog/advertisements/advertisements.py
```diff
@@ -25,6 +25,10 @@ class AdvertisementsResource(Resource):
         },
         'meta': {
             'type': 'dict',
+            'mapping': {
+                'type': 'dict',
```
petrjasek: in the mapping this type should be object we send it to elastic as it is, and if you want this not to be indexed at all, you should use `'enabled': false` rather than `not_analyzed` which avoid analyzer but indexes the whole thing.
[more info](https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html)

PR #986, opened 2017-11-08
Style fix

Aca-jov:


server/liveblog/themes/themes_assets/default/templates/template-post.html
```diff
@@ -1,7 +1,7 @@
 {% macro translate(text) %}{% if not options.i18n[settings.language] %}{{ text }}{% elif not options.i18n[settings.language][text] %}{{ text }}{% else %}{{ options.i18n[settings.language][text] }}{% endif %}{% endmacro %}
 
 {% set mainItem = item.groups[1].refs[0].item %}
-{% set postDate = item.published_date %}
+{% set postDate = item.content_updated_date %}
```
nistormihai: why? this is this is contrary to:
https://dev.sourcefabric.org/browse/LBSD-1935

Aca-jov: I think its not contrary to that, because content_update_date is equal to published date (when item is published for the first time) and later when item is updated, it shows updated time (which is required by https://dev.sourcefabric.org/browse/LBSD-1845).

nistormihai: LBSD-1845 states to add the `Update date field` not to change the current behavior of the published date, the main date showed.

Aca-jov: There was a bit misunderstanding in ticket descripion, and comment by backo. But I see it is switched to 3.3.2, so not merging this.

PR #994, opened 2017-11-16
Themes update

tomaszrondio:


tomaszrondio:
I know these commit messages are ugly but I cannot rebase, squash and even rename without conflicts with angular theme O_o

PR #1007, opened 2017-12-06
Fix/elastic search mappings

nistormihai:


scanterog:
LGMT. Tested locally and working fine.

PR #1014, opened 2017-12-11
force fb, tw, instagram, youtube embeds to be embedded via embedly

tomaszrondio:


client/app/scripts/sir-trevor-blocks/index.js
```diff
@@ -128,7 +156,7 @@ angular
                     var input = $(this)
                         .text()
                         .trim();
-
+
```
nistormihai: pls remove this.

PR #1029, opened 2018-01-11
Fix: Keep theme settings when updating the instance

GyanP:


server/liveblog/themes/themes.py
```diff
@@ -435,29 +435,9 @@ def _save_theme_settings(self, theme, previous_theme):
 
         # Check if theme settings are changed.
         if 'settings' in previous_theme:
-            options = []
-            old_theme_settings = {}
```
jey07ro: new defaults should be reflected in previous "defaulted" settings, so we should keep this code,
but change to support new settings also.

server/liveblog/themes/themes.py
```diff
@@ -456,8 +456,8 @@ def _save_theme_settings(self, theme, previous_theme):
                 else:
                     # Otherwise we keep the settings that are already on the theme.
                     default_theme_settings[key] = default_prev_theme_settings[key]
-
             theme_settings.update(default_theme_settings)
+            theme_settings.update(default_prev_theme_settings)
```
nistormihai: new settings and new defaults will now be updated because it is updating with the previous defaults.

PR #1030, opened 2018-01-11
Feature: Sets correct width and height to amp-iframes to stay compatible

eos87:


eos87:
Hi @nistormihai, branch has been updated :)

PR #1033, opened 2018-01-16
Warn the user of unsaved content when exiting the current blog

tomaszrondio:
https://dev.sourcefabric.org/browse/LBSD-2036


nistormihai:
please adjust the e2e test for this one.

PR #1036, opened 2018-01-19
Fix/error published scorecard

GyanP:
https://dev.sourcefabric.org/browse/LBSD-2051
https://dev.sourcefabric.org/browse/LBSD-2008


nistormihai:
eslint issue

PR #1043, opened 2018-01-24
Add support for relative date format

mstoltenburg:
I want to support relative dates in AMP templates too, so the idea is to enable something like `{{ postDate | date('ago') }}`. You already support this format for templates parsed by nunjucks, see https://github.com/liveblog/liveblog/blob/9c134ec1538f3fd58050c043d7cea44f8bb9471e/server/liveblog/themes/themes_assets/default/js/theme/helpers.js#L13-L15

### DIFF
--- a/server/liveblog/themes/template/filters.py
+++ b/server/liveblog/themes/template/filters.py
@@ -16,6 +16,7 @@ def moment_date_filter_container(theme):
         :return: str
         """
         settings = theme.get('settings', {})
+        locale = settings.get('language', 'en')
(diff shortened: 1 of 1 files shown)


nistormihai:
we only added the ago timestamp just to have the same format as the javascript when it loads.
But time ago doesn't work on server side, because isn't generated each second or minute is generated when  a new post  is pushed. for amp time ago we need something like
https://www.ampproject.org/docs/reference/components/amp-timeago
but this PR doesn't hurt.

PR #1058, opened 2018-02-02
Fix: Introduced a mechanism to only activate the publish button, when…

GyanP:
… the upload has been completed
https://dev.sourcefabric.org/browse/LBSD-2053?filter=-1


nistormihai:
All good

PR #1062, opened 2018-02-12
Fix: created a mechanism to remove uploaded image when creating bog

GyanP:
https://dev.sourcefabric.org/browse/LBSD-1975


client/app/scripts/liveblog-bloglist/views/main.html
```diff
@@ -204,6 +204,7 @@ <h3 translate>Create New Blog</h3>
                                         </div>
                                     </section>
                                     <p class="warn-block">If you are using the Liveblog 3 AMP theme, please make sure to use an aspect ratio of 3x2 (600x400 pixels) for the blog image.</p>
+                                    <a href="" class="btn btn-default pull-right" ng-click="removeImage()" translate>Remove</a>
```
nistormihai: please add this only when an image exists,
also better use a button instead of a link.

PR #1068, opened 2018-02-23
Fix: fixed duplicate users issue on Create blog and settings and filt…

GyanP:
…ered admin from dropdown
https://dev.sourcefabric.org/browse/LBSD-2066


client/app/scripts/liveblog-bloglist/directives/user-select-list.js
```diff
@@ -101,6 +108,11 @@ export default function lbUserSelectList(api) {
 
             scope.choose = function(user) {
                 scope.onchoose({user: user});
+                var exists = scope.members.some((el) => el._id === user._id);
```
nistormihai: good usage

PR #1070, opened 2018-02-27
feature: ads not rotating for default theme

eos87:
Hi @nistormihai, please have a look when you have some time


server/liveblog/advertisements/advertisements.py
```diff
@@ -60,3 +67,33 @@ def on_updated(self, updates, original):
...
+            output['collection'] = collection
+
+    if output and output.get('collection', False):
+        ads_ids = output['collection'].get('advertisements', [])
+        ads_ids = list(map(lambda x: x['advertisement_id'], ads_ids))
+
+        ads_query = get_resource_service('advertisements').find({"_id": {"$in": ads_ids}})
+        ads = list(ads_query)
+
+        # looping over them, not sure which one is more performant
```
nistormihai: please remove comments.

server/liveblog/advertisements/advertisements.py
```diff
@@ -60,3 +67,33 @@ def on_updated(self, updates, original):
...
+        else:
+            collection = get_resource_service('collections').find_one(req=None, _id=output.get('collection'))
+            output['collection'] = collection
+
+    if output and output.get('collection', False):
+        ads_ids = output['collection'].get('advertisements', [])
+        ads_ids = list(map(lambda x: x['advertisement_id'], ads_ids))
+
+        ads_query = get_resource_service('advertisements').find({"_id": {"$in": ads_ids}})
+        ads = list(ads_query)
```
nistormihai: shouldn't ads be defined empty list outside of the if ?
so `return api_response(ads, 200)`  will not break if no ads defined.

eos87: Yeah, you're right. It will break if `if output and output.get('collection', False)` evaluate to False. Fixed

PR #1073, opened 2018-03-01
Feature/brightcove support

nistormihai:


server/liveblog/themes/template/filters.py
```diff
@@ -71,7 +70,20 @@ def addten(date):
...
+        '(?P<account>\d*)/',
+        '(?P<player>\w*)',
+        '_',
+        '(?P<embed>\w*)',
+        '\/index\.html\?videoId=',
+        '(?P<videoId>\d*)'
+    ]
+    match = re.compile(''.join(brightcove_re_array)).match(html)
+    if match:
+        return ('<amp-brightcove data-account="{account}" data-player="{player}" data-embed="{embed}" data-video-id="{videoId}" \
```
nistormihai: it is much more easy like this but it isn't flexible the `width="480" height="270"` are set from the filter and not on the theme level.

nistormihai: even if it is easy like that we need to provide the flexibility to change this on the theme level. please adjust this as it is done with twitter Facebook and so on ... the only big problem it is with passing multiple meta data instead just one `original_id`

server/liveblog/themes/template/filters.py
```diff
@@ -71,7 +70,20 @@ def addten(date):
 
 
 def ampify(html):
-    if re.search('iframe', html, re.IGNORECASE):
+    brightcove_re_array = [
```
nistormihai: the string is to big to pass `flake8` and spiting into the next line doesn't look so good
so list that is joined into a string looks much better IMHO

server/liveblog/themes/template/filters.py
```diff
@@ -71,7 +70,20 @@ def addten(date):
...
+        'https?://(www\.)?players\.brightcove\.net/',
```
nistormihai: the issue with the previous string was that it didn't started with proper url.`https?://(www\.)?`

nistormihai:
@GyanP `brightcove_re_array` is just your version of regex
The changes are just a quick fix please check the proper fix as stated in the above comments.

PR #1074, opened 2018-03-04
Fix/notification

nistormihai:


gideonlehmann:
Fix works for inviting team members in the blog creation modal and via blog settings. It fails though when someone requests access to the blog and is accepted by the blogowner. I reopened the ticket.

PR #1076, opened 2018-03-05
Fix refreshing key syndication warning

GyanP:
https://dev.sourcefabric.org/browse/LBSD-1880


server/liveblog/themes/themes_assets/default/theme.json
```diff
@@ -462,4 +462,4 @@
           "credit:": "credit:"
         }
     }
-}
+}
```
nistormihai: please remove this from the commit, not relevant.

GyanP: It is a commit from devel branch which came to feature branch by its reference. and its very old commit to devel so, for now, I do not think it makes sense to do this. but for future, I will keep in mind for sure

PR #1082, opened 2018-03-07
Fix/amp theme icons

GyanP:
https://dev.sourcefabric.org/browse/LBSD-2118
https://dev.sourcefabric.org/browse/LBSD-2119


aleksandarjelicic:
@tomaszrondio will take over, we have a problem with new version of the css that got overwritten.

PR #1092, opened 2018-03-12
feature: LBSD-2122 additional scroll to element on window load

tomaszrondio:


server/liveblog/themes/themes_assets/default/js/theme/view.js
```diff
@@ -351,6 +351,10 @@ function permalinkScroll() {
   if (scrollElem) {
     scrollElem.classList.add('lb-post-permalink-selected');
     scrollElem.scrollIntoView();
+    window.onload = function() {
+      scrollElem.scrollIntoView();
+      console.log("onload", scrollElem);
```
nistormihai: do we need console.log here?

PR #1102, opened 2018-03-15
feature: auto updating relative date on posts list

tomaszrondio:
LBSD-2120


nistormihai:
you have some eslint issues except that it looks good.

PR #1103, opened 2018-03-16
Fix/image upload fails

GyanP:
https://dev.sourcefabric.org/browse/LBSD-2103


nistormihai:
please check this PR out there are a lot of commits that I don't think belong here
also what is the difference from https://github.com/liveblog/liveblog/pull/1104 ?

GyanP:
I think they all commit because I have updated the respective branch from devel before making changes. and also #1103  and #1104  both are different tasks. #1103  have changes in "client/app/scripts/liveblog-edit/directives/freetype-image.js" and #1104  have changes in "client/app/scripts/liveblog-bloglist/directives/user-select-list.js"

PR #1122, opened 2018-04-10
Simple theme

aleksandarjelicic:
@nistormihai This are files for new child simple theme. Please register this theme so it becomes system theme, and available in the theme settings.


aleksandarjelicic:
I get {"_message": "File not found on media storage.", "_status": "ERR"} on liveblog output.

nistormihai:
please run `gulp template-inject` and then commit `template.html`

PR #1133, opened 2018-04-20
feature: bulk action archive and delete for blogs

GyanP:
https://dev.sourcefabric.org/browse/LBSD-2155


client/app/scripts/liveblog-bloglist/controllers/blog-list.js
```diff
@@ -87,6 +90,32 @@ export default function BlogListController(
...
+        }, () => {
+            notify.pop();
+            notify.error(gettext('Something went wrong'));
+        });
+    }
+
+    $scope.askRemoveBlog = function() {
+        modal.confirm(gettext('Are you sure you want to delete the blog(s)?'))
+            .then(() => {
+                var selectedBlogs = new Array();
```
nistormihai: why `new Array()` ? this cause extra processing on runtime

PR #1145, opened 2018-05-05
LBSD-2173 fix: adding the GA tracking code to a theme in the theme settings does not work

eos87:
After figuring out this and testing with GA account, I ended up using same approach and just making a really dummy change. 

p.s. To check this is working I used chrome extensions: GA Debug and Tag Assistant


eos87:
I won't put more time into this one until next sprint. Closing.

PR #1151, opened 2018-05-10
LBSD-2116 fix: decreased versions for nunjucks packages

eos87:
Since more recent versions of nunjucks use beta versions of babel packages, it breaks the making of new theme
extensions given that we're using stable v6 of babel. Reducing the version of nunjucks packages solves the problem


server/liveblog/themes/themes_assets/default/gulpfile.js
```diff
@@ -2,7 +2,6 @@
 
 const gulp = require('gulp');
 const browserify = require('browserify');
-const nunjucksify = require('nunjucksify');
```
tomaszrondio: why is this removed?

eos87: It's not required anymore

PR #1154, opened 2018-05-11
Fixed issue with automation test for blog settings

GyanP:
Blog settings automation test


eos87:
I think this should be :nth-child(4)

tomaszrondio:
this is crazy. Lets get rid of it like you said and put display none inline

PR #1159, opened 2018-05-15
fix: added version, this need to be increased.

nistormihai:


client/app/scripts/sir-trevor-blocks/index.js
```diff
@@ -234,6 +234,7 @@ angular
                     description: self.$('.description-preview').text(),
                     credit: self.$('.credit-preview').text(),
                     syndicated_creator: this.getData().syndicated_creator,
+                    liveblog_version: '3.4',
```
nistormihai: I'm not sure this is well put, a different place where all the clients can use it, or extracted from package.json.

PR #1165, opened 2018-05-21
Travis yml setting to run down stream build

GyanP:
https://dev.sourcefabric.org/browse/LBSD-2112


eos87:
Hi @GyanP, do you think we should better put it on `after_success` section?

PR #1180, opened 2018-06-06
Feature/outputchannel editors

GyanP:
https://dev.sourcefabric.org/browse/LBSD-2231

### DIFF
--- a/server/liveblog/prepopulate/app_initialize.py
+++ b/server/liveblog/prepopulate/app_initialize.py
@@ -36,7 +36,7 @@ dictionary:
 """
 __entities__ = OrderedDict([
     ('global_preferences', ('global_preferences.json', ['key'], False)),
-    ('roles', ('roles.json', ['name'], False)),
+    ('roles', ('roles.json', ['name'], True)),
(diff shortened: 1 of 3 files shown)


server/liveblog/prepopulate/app_initialize.py
```diff
@@ -36,7 +36,7 @@
 """
 __entities__ = OrderedDict([
     ('global_preferences', ('global_preferences.json', ['key'], False)),
-    ('roles', ('roles.json', ['name'], False)),
+    ('roles', ('roles.json', ['name'], True)),
```
eos87: Hi @GyanP, just out of curiosity. I haven't checked this part of the code. What does it mean the True or False value at the end of the tuple?

GyanP: It responsible for `do_patch` parameter's value in import_file (https://github.com/liveblog/liveblog/blob/devel/server/liveblog/prepopulate/app_initialize.py#L247) which is written in superdesk and `do_patch` is mainly a check to override the values from the roles.json file or not

PR #1181, opened 2018-06-06
fix: don’t load deleted posts when toggling highlighted

### LINKED ISSUE #1179: Clicking the "Highlights" button shows deleted posts
Whenever you use the "Highlights" button in the default theme, all deleted posts show up, even if you uncheck the option again.

We should set `opts.notDeleted = true;` here like we do in `vm.loadPostsPage`.
https://github.com/liveblog/liveblog/blob/d0cbaaa29ca9c139892b80ed96899c93a2e3c66e/server/liveblog/themes/themes_assets/default/js/theme/viewmodel.js#L152

It's only set correctly for `loadSort` at https://github.com/liveblog/liveblog/blob/d0cbaaa29ca9c139892b80ed96899c93a2e3c66e/server/liveblog/themes/themes_assets/default/js/theme/handlers.js#L141

mstoltenburg:
my try to fix #1179

### DIFF
--- a/server/liveblog/themes/themes_assets/default/js/theme/handlers.js
+++ b/server/liveblog/themes/themes_assets/default/js/theme/handlers.js
@@ -90,7 +90,7 @@ var buttons = {
 
       highlightButton.classList.toggle('header-bar__highlight--active');
       LB.settings.onlyHighlighted = !LB.settings.onlyHighlighted;
-      return viewmodel.loadPosts()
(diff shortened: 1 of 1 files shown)


tomaszrondio:
because of not loading deleted posts from API liveblog embed won't get notified that post was deleted. Therefore it won't remove it from timeline. Solution should be on template level. Posts with `deleted: true` should not be rendered.

PR #1190, opened 2018-06-14
fix: outputchannel creation when style_tag is none

GyanP:
fix when creating output channel when ESI settings is on


eos87:
Closing since we can merge master back into devel once release is done

PR #1210, opened 2018-07-11
fix: Ingested channel has to be revisited in order for reposts to appear

GyanP:
https://dev.sourcefabric.org/browse/LBSD-2236


GyanP:
Hi @eos87, yes it can be reviewed, task's link is attached with PR description https://dev.sourcefabric.org/browse/LBSD-2236 where you can find its steps to reproduce/test

PR #1266, opened 2019-09-05
cross-plattform path seperator in template.js

cbieser:
Fixed cross-platform path separator problem


eos87:
Hi @cbieser. This PR addressed something different to #1267. Correct?

cbieser:
Hi @eos87 after fixing the path seperator local #1267 was the next error. Currently I use local data i've exported from the API to the `test/api_response.json` and `text/options.json` files.

eos87:
I see. I changed the base branch as you may notice because `v3.6.7` is the upcoming release. I'll give a look at #1267 as soon as I have some time to. Thanks for this PR.

PR #1284, opened 2020-04-15
LBSD-2489 Improve Liveblog readme file and installation steps

dev-tsoni:
**Updates**:
- Updated Dockerfile along with bionic base image
- Updated README.md with updated installation instructions using docker


docker/common.yml
```diff
@@ -59,6 +59,18 @@ superdesk:
    - SUPERDESK_WS_URL=ws://127.0.0.1/ws
    - SUPERDESK_CLIENT_URL=http://127.0.0.1
    - SUPERDESK_TESTING=True
+   - EMBED_PROTOCOL=http://
+   - LIVEBLOG_DEBUG=False
+   - SYNDICATION=True
+   - MONGO_DBNAME=superdesk
+   - SUBSCRIPTION_LEVEL=network
+   - LIVEBLOG_CACHE_TYPE=redis
+   - LIVEBLOG_CACHE_REDIS_URL=redis://redis:6379/12
```
eos87, line 68: Let's remove the redis cache settings, by default in local env the `simple` cache is easier approach.

docker/common.yml
```diff
@@ -59,6 +59,18 @@ superdesk:
...
+   - BOTO_CONFIG=/doesnotexist
+   - SYNDICATION_VALIDATE_AUTH=0
+   - MAIL_SUPPRESS_SEND=False
+   - EMBEDLY_KEY=82645d4daa7742cc891c21506d28235e
```
eos87, line 72: I'm wondering how could we avoid passing this embody key. It's a private key in the end and we should not expose either our own key or other random key. Also I'm not sure of the legal implications. Let's better remove it to leave it somehow indicated that needs to be replaced.

docker/common.yml
```diff
@@ -59,6 +59,18 @@ superdesk:
    - SUPERDESK_WS_URL=ws://127.0.0.1/ws
    - SUPERDESK_CLIENT_URL=http://127.0.0.1
    - SUPERDESK_TESTING=True
+   - EMBED_PROTOCOL=http://
+   - LIVEBLOG_DEBUG=False
+   - SYNDICATION=True
+   - MONGO_DBNAME=superdesk
```
eos87, line 65: let's change the DBNAME to liveblog instead of superdesk

docker/common.yml
```diff
@@ -59,6 +59,18 @@ superdesk:
    - SUPERDESK_WS_URL=ws://127.0.0.1/ws
    - SUPERDESK_CLIENT_URL=http://127.0.0.1
    - SUPERDESK_TESTING=True
+   - EMBED_PROTOCOL=http://
+   - LIVEBLOG_DEBUG=False
```
eos87, line 63: It may be better to have LIVEBLOG_DEBUG=True.

dev-tsoni:
Made the requested changes and changed the folder name to server instead of liveblog

docker/common.yml
```diff
@@ -59,6 +59,18 @@ superdesk:
    - SUPERDESK_WS_URL=ws://127.0.0.1/ws
    - SUPERDESK_CLIENT_URL=http://127.0.0.1
    - SUPERDESK_TESTING=True
+   - EMBED_PROTOCOL=http://
+   - LIVEBLOG_DEBUG=True
+   - SYNDICATION=True
+   - MONGO_DBNAME=liveblog
+   - SUBSCRIPTION_LEVEL=network
+   - LIVEBLOG_CACHE_TYPE=simple
+   - LIVEBLOG_CACHE_REDIS_URL=redis://redis:6379/12
```
eos87, line 68: One last comment. Now that `LIVEBLOG_CACHE_TYPE=simple` we don't need the `LIVEBLOG_CACHE_REDIS_URL ` so please just remove it. Thanks

Dockerfile
```diff
@@ -24,11 +20,10 @@ ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
 
 # setup the environment
-WORKDIR /opt/superdesk/
+WORKDIR /opt/server/
 COPY ./docker/nginx.conf /etc/nginx/nginx.conf
 COPY ./docker/superdesk_vhost.conf /etc/nginx/sites-enabled/superdesk.conf
 COPY ./docker/start.sh /opt/superdesk/start.sh
```
eos87, line 26: @dev-tsoni will this `/opt/superdesk/start.sh` after renaming the folder to `server/`?

Dockerfile
```diff
@@ -41,22 +36,26 @@ EXPOSE 5100
...
 # copy client sources
-COPY ./client /opt/superdesk/client
+COPY ./client ./client
 
 # TODO: this is hack to update basic themes during bamboo deployment
-COPY ./server/liveblog/themes/themes_assets/ /opt/superdesk/themes_assets/
+COPY ./server/liveblog/themes/themes_assets/ ./themes_assets/
 
-RUN cd ./client && npm i
+CMD /opt/superdesk/start.sh
```
eos87, line 61: this should be `/opt/server/start.sh`? (because of folder rename)

docker/common.yml
dev-tsoni: ok, will remove LIVEBLOG_CACHE_REDIS_URL

Dockerfile
dev-tsoni: ah yes, will move the start script to server

PR #1290, opened 2020-05-12
LBSD-2492 Update Live Link

dev-tsoni:
Updates:
- Hide Live link until public_url is received
- Use EMBED_PROTOCOL for setting Live link's protocol

### DIFF
--- a/client/app/scripts/liveblog-edit/blog.service.js
+++ b/client/app/scripts/liveblog-edit/blog.service.js
@@ -30,7 +30,7 @@ export default function blogService(api, $q, $rootScope, config) {
     * @param {object} blog
     * @return {promise} public_url
     **/
-    function getPublicUrl(blog) {
+    function getPublicUrl(blog, $scope = {}) {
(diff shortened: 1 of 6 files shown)


client/app/scripts/liveblog-edit/blog.service.js
```diff
@@ -30,7 +30,7 @@ export default function blogService(api, $q, $rootScope, config) {
     * @param {object} blog
     * @return {promise} public_url
     **/
-    function getPublicUrl(blog) {
+    function getPublicUrl(blog, $scope = {}) {
```
eos87, line 33: I think passing the `$scope` as parameter this way is not a good practice. Can you please explain the reason of this? Can we improve it? Thanks

dev-tsoni: The reason was to update the pulicUrl attribute in the scope when the publicUrl arrives. Okay, I will improve it.

PR #1293, opened 2020-05-19
LBSD-2510 Embed enhancer script

eos87:


dev-tsoni:
Works fine and looks good to me.

PR #1294, opened 2020-05-19
LBSD-2487 Ad collection rotation improvements

eos87:
Some users have requested to avoid having the first advertisement at the top of the timeline. So I improved this in Default SEO and AMP which are the ones that support ads injection. To test, please create an output channel and assign a group of ads (you can create the ads in the Ad Manager panel). Thanks for checking.


dev-tsoni:
Looks good to me

PR #1295, opened 2020-05-21
LBSD-2479 Resets comment form after submit to allow sending another without reloading browser

eos87:
The expected behavior for this task was not possible because of AMP limitations but I took advantage of it to improve AMP either way.

To test: 
- Create a blog and set AMP Theme
- Allow comments in theme settings
- Send a comment and try to send another right away

Thanks for checking.


dev-tsoni:
Works fine!

PR #1354, opened 2023-12-08
Removed node-sass for sass

BrianMwangi21:
Resolves LBSD-2757

Replacing node-sass with a better SASS processing alternative in Liveblog


eos87 (approved):
I tested this locally and works well. As we rely on `super desk-client-core`, we might provide a PR to move that one away from `node-sass` as it is also deprecated.

PR #1356, opened 2023-12-20
[LBSD-2756] Resolves sharing of post on social media

BrianMwangi21:
Resolves [LBSD-2756] 

When a post is shared and the link opened, it shows the post shared first, with previous posts following and a counter for new updates - if they exist.

[LBSD-2756]: https://sofab.atlassian.net/browse/LBSD-2756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BrianMwangi21:
Screenshot of shared post appearing as the first and the update modal with count above it
image

server/liveblog/themes/themes_assets/default/js/theme/view.js
```diff
@@ -147,17 +147,27 @@ function addPosts(posts, position) {
...
+  }
+
   if (pending.length === 1) {
-    one.toggle('mod--displaynone', false);
-    updates.toggle('mod--displaynone', true);
+    updateToggles(false, true, true);
   } else if (pending.length > 1) {
-    one.toggle('mod--displaynone', true);
-    updates.toggle('mod--displaynone', false);
+    if(permalink) {
```
eos87, line 163: nit:
```suggestion
    if (permalink) {
```

server/liveblog/themes/themes_assets/default/js/theme/viewmodel.js
```diff
@@ -268,15 +272,37 @@ vm.init = function() {
...
+    .catch(error => console.log(error));
   }
 
   var isBlogOpen = LB.blog.blog_status === "open";
   var tenSeconds = 10 * 1000;
 
   if (isBlogOpen) {
-    // let's hit backend right away after load and render latest updates
-    fetchLatestAndRender();
+    if(permalink._id) {
```
eos87, line 292: nit:
```suggestion
    if (permalink._id) {
```

server/liveblog/themes/themes_assets/default/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "liveblog-default-theme",
-  "version": "3.6.68",
+  "version": "3.6.77",
```
eos87, line 3: I think once this functionality is fully ready we should bump this to be a minor version, like `3.7.0`, it's ok for now though

server/liveblog/themes/themes_assets/default/theme.json
```diff
@@ -1,7 +1,7 @@
 {
     "label": "Liveblog 3 Default SEO Theme",
     "name": "default",
-    "version": "3.6.68",
+    "version": "3.6.77",
```
eos87, line 4: Same as my previous comment :)

eos87 (changes requested):
It seems like the references to the build files are broken and not sync between the manifest and the theme file. See my screenshots

This is how it looks when I open a blog in the embed endpoint:
image

and here you have the manifest and theme json files:
image

image

I'm not fully sure why this is happening but perhaps you're using `gulp-watch` without `gulp`? Please double check it. You will need to fix these references manually in both files to point to the right build and then bundle again with gulp maybe.

server/liveblog/themes/themes_assets/default/js/theme/view.js
BrianMwangi21: Resolved.

server/liveblog/themes/themes_assets/default/js/theme/viewmodel.js
BrianMwangi21: Resolved.

BrianMwangi21:
Style and scripts now match in rev-manifest and theme.json

BrianMwangi21 (commented):
Requesting review for new changes,

server/liveblog/themes/themes_assets/default/js/theme/viewmodel.js
```diff
@@ -262,24 +310,21 @@ vm.init = function() {
...
-      tags: selectedTags
-    })
-    .then(view.renderPosts)
-    .then(view.initGdprConsentAndRefreshAds);
-  }
-
   var isBlogOpen = LB.blog.blog_status === "open";
   var tenSeconds = 10 * 1000;
 
   if (isBlogOpen) {
```
eos87, line 316: I think the sharing logic should work also if the blog is no longer open (archived).

client/app/scripts/liveblog-embed-script/sharing.ts
```diff
@@ -1,71 +1,43 @@
-import modalHtml from './ui/shared-post-modal.html';
```
eos87, line 1: please also delete the `./ui/shared-post-modal.html` template

eos87 (changes requested):
After clicking in the alert "X new updates" the reddit embedded post does not render. Here it is an example below: 
image

Also, when you clicked a shared post and then, instead of clicking the "X new updates" alert you change the order, all the posts are shown but, the alert did not disappear:
image

I left a few comments, please take a look. We're getting there! 

Thanks!

server/liveblog/themes/themes_assets/default/theme.json
```diff
@@ -1,7 +1,7 @@
 {
     "label": "Liveblog 3 Default SEO Theme",
     "name": "default",
-    "version": "3.6.68",
+    "version": "3.6.187",
```
eos87, line 4: This version should match the one in package.json file

eos87 (changes requested):
There is one last comment I left regarding version. Please check it 🙏

eos87 (approved):
all good, well done!

PR #1358, opened 2024-01-31
[LBSD-2763] - Create the UI components for the Editor's interface

BrianMwangi21:
This PR introduces the implementation of static UI components for the Editor's interface. Each component has been developed according to the specifications and designs provided in the attached design documents for this ticket. The goal is to lay down the foundational UI elements that will be utilized in the Editor's interface.

Screenshots:
image
image


eos87 (changes requested):
Let's change the components to be function components instead of class ones. We should also have a separate talk about the naming conventions we're going to use to have more consistency in the project.

client/app/scripts/liveblog-edit/components/polls/poll-component-create.tsx
```diff
@@ -0,0 +1,74 @@
...
+                                <span
+                                    className="poll_option_remove_container"
+                                    onClick={() => removeOption(index)}
+                                >
+                                    <span className="icon-close-small" />
+                                </span>
+                            </div>
+                        );
+                    })}
+                    <p className="poll_option_add" onClick={addOption}>+ Add Option</p>
```
eos87, line 45: Semantically speaking, we should use either an `a` tag or a `button` tag here instead of a `p`.

client/app/scripts/liveblog-edit/components/polls/poll-component-view.tsx
```diff
@@ -0,0 +1,63 @@
+/* eslint camelcase: "off" */
+import React, { useState, useEffect } from 'react';
+import './poll-component.scss';
+import { pollCalculations } from './utils';
+
+export interface IPollBody {
+    active_until: string;
```
eos87, line 7: I suppose the reason for this underscore case it's because it comes from backend?

eos87 (changes requested):
This doesn't look good, the icons should fit in one line, as it looks in the design
image

--------------------------
When clicking "Reset Poll" button, show this error which breaks the UI 
image

I also left a couple of other minor comments, please check.

client/app/scripts/liveblog-edit/components/polls/poll-component-view.tsx
BrianMwangi21: Yes that's the case.

eos87 (approved):
Looks good to me!

PR #1357, opened 2024-01-29
[LBSD-2762] - Implement backend logic to store and manage polls in the Editor's interface

BrianMwangi21:
This PR introduces the functionality of incorporating polls into posts as a new variation of items that can be used. A summary of the key objectives and the implemented solutions are as follows:

#### Database Collection for Polls:
A new collection named `polls_collection` has been added to our database. This collection is designed to store poll data, including questions, options, and votes, allowing for dynamic interaction within blogs.

#### CRUD Endpoints for Polls:
Implementation of new endpoints to support Create, Read, Update, and Delete (CRUD) operations for polls. These endpoints facilitate the management of poll data, enabling users to easily create new polls, view existing polls, update poll details, and remove polls when necessary.

#### Integration of Polls into Post:
The existing endpoint for post items has been adjusted to include the functionality of attaching polls. This adjustment allows polls to be seamlessly integrated into posts.

#### Conditional Operations Based on Poll Attachment:
To ensure the integrity of data and interactions, we've implemented conditional logic within the CRUD actions for post items. This logic checks for the attachment of a poll to a post item and performs the corresponding operation.

PR #1360, opened 2024-02-16
[LBSD 2764] - Connect the backend with frontend components for Polls feature

BrianMwangi21:
### Purpose
This PR aims to integrate the UI components for the Editor's interface, developed in [LBSD-2763](https://github.com/liveblog/liveblog/pull/1358), with the backend endpoints implemented in [LBSD-2762](https://github.com/liveblog/liveblog/pull/1357). The goal is to enable the storage and management of polls within the Editor's interface.

### What has changed
- Wired the frontend UI components for polls with the backend endpoints to allow poll management.
- A poll can be created as a single item in a post or be one of several items in a post.
- Enhanced the "delete" warning for items and posts to inform users that deleting an item containing a poll will result in the loss of all associated poll data.

### Steps to test

3. The published poll can also be highlighted, pinned, removed, updated, unpublished and saved in contributions and also moved around to change the order.
4. A new poll can also be saved as a draft or scheduled for publishing at a later date.
5. To test the removal warning, click the remove icon on a published post to see the warning.

Resolves: [LBSD-2764]

### DIFF
--- a/client/app/scripts/liveblog-edit/posts.service.ts
+++ b/client/app/scripts/liveblog-edit/posts.service.ts
@@ -307,17 +307,31 @@ const postsService = (api, $q, _userList, session) => {
 
             // save every items
             _.each(items, (itemParam) => {
-                const item = {
-                    blog: blogId,
-                    text: itemParam.text,
-                    meta: itemParam.meta,
-                    group_type: itemParam.group_type,
-                    item_type: itemParam.item_type,
-                    commenter: itemParam.meta && itemParam.meta.commenter,
-                    syndicated_creator: itemParam.syndicated_creator,
-                };
+                switch (itemParam.item_type) {
+                case 'poll': {
+                    const poll = {
+                        blog: blogId,
+                        poll_body: itemParam.poll_body,
+                    };
 
-                dfds.push(api.items.save(item));
+                    dfds.push(api.polls.save(poll));
+                    break;
+                }
+                default: {
+                    const item = {
+                        blog: blogId,
+                        text: itemParam.text,
+                        meta: itemParam.meta,
+                        group_type: itemParam.group_type,
+                        item_type: itemParam.item_type,
+                        commenter: itemParam.meta && itemParam.meta.commenter,
+                        syndicated_creator: itemParam.syndicated_creator,
+                    };
+
+                    dfds.push(api.items.save(item));
+                    break;
+                }
+                }
             });
         }
 
@@ -341,7 +355,14 @@ const postsService = (api, $q, _userList, session) => {
 
                 // update the post reference (links with itemsList)
                 _.each(itemsList, (item) => {
-                    post.groups[1].refs.push({ residRef: item._id });
+                    switch (item.item_type) {
+                    case 'poll':
+                        post.groups[1].refs.push({ residRef: item._id, location: 'polls' });
+                        break;
+                    default:
+                        post.groups[1].refs.push({ residRef: item._id });
+                        break;
+                    }
                 });
             }
 
--- a/server/liveblog/posts/posts.py
+++ b/server/liveblog/posts/posts.py
@@ -420,12 +420,34 @@ class PostsService(ArchiveService):
             content_diff = True
         else:
             for index, val in enumerate(updates["groups"][1]["refs"]):
-                item = get_resource_service("archive").find_one(
-                    req=None, _id=val["residRef"]
-                )
-                if item["text"] != original["groups"][1]["refs"][index]["item"]["text"]:
-                    content_diff = True
-                    break
+                service_name = val.get("location", "archive")
+                service = get_resource_service(service_name)
+                item = service.find_one(req=None, _id=val["residRef"])
+                item_type = item.get("item_type")
+
+                if item_type == "poll":
+                    item_poll_body = item.get("poll_body", {})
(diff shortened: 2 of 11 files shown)


client/app/scripts/liveblog-edit/posts.service.ts
```diff
@@ -307,17 +307,31 @@ const postsService = (api, $q, _userList, session) => {
...
-
-                dfds.push(api.items.save(item));
+                switch (itemParam.item_type) {
+                case 'poll': {
+                    const poll = {
+                        blog: blogId,
+                        poll_body: itemParam.poll_body,
+                    };
+
+                    dfds.push(api.polls.save(poll));
```
eos87, line 317: I know this variable `dfds` was already there but, could you please rename to something that has some meaning?

server/liveblog/posts/posts.py
```diff
@@ -420,12 +420,45 @@ def on_update(self, updates, original):
             content_diff = True
         else:
             for index, val in enumerate(updates["groups"][1]["refs"]):
```
eos87, line 422: nit:
```suggestion
            for index, item_ref in enumerate(updates["groups"][1]["refs"]):
```

server/liveblog/posts/posts.py
```diff
@@ -420,12 +420,45 @@ def on_update(self, updates, original):
...
-                    content_diff = True
-                    break
+                service_name = val.get("location", "archive")
+                service = get_resource_service(service_name)
+                item = service.find_one(req=None, _id=val["residRef"])
+                item_type = item.get("item_type")
+
+                if item_type == "poll":
+                    item_poll_body = item.get("poll_body", {})
+                    original_poll_body = original["groups"][1]["refs"][index][
```
eos87, line 430: this piece of code is used multiple times and it seems to me that we get it from the `enumerate`. Do you know what's the reason to get it like again and again? See line `458` too. 
`original["groups"][1]["refs"][index]`
If it's not possible to use the `item_ref` suggested above from the `enumerate`, perhaps we should extract it once and reuse the variable, I think that would make the code more readable.

eos87: Another thing is that all these code from line `415` to `461` is to detect if the content has changed. Maybe we should extract it into a separate function and cover it with tests. What do you think?

eos87 (changes requested):
I left a few minor comments. Please check. Thanks!

server/liveblog/posts/posts.py
BrianMwangi21: I have extracted the section into a function called `check_content_diff`. 

Also note that I have simplified the check for polls to just check the `active_until` value since that's the only value that can be changed from the Editor's end before the poll elapses. Hence the question and answers aren't checked.

eos87 (changes requested):
Hi @BrianMwangi21, I have a few comments and suggestions I would like you to check:

Polls should allow only hours when creating and editing
allow-zero-days-only-hours

This text should be "edit" when editing the post to reflect the right state
should-be-update

The time left is not being reflected properly. If there are only hours, it should be hours, or minutes. I think `moment.js` has a function for that which is called `.fromNow()` if I'm not mistaken.
should-reflect-hours-left

When the post is edited, it seems like the polls are being duplicated and we're quickly creating garbage in the DB. I think this happens with other items as well but we should prevent this with our new features. Also, when the blog is removed, these polls are being left there
every-update-garbage-left

Let me know if you need to talk about any of these suggestions.

eos87 (changes requested):
Hi @BrianMwangi21, there is another problem now. When you update a post with a poll and save it, the poll disappears. Please check it

BrianMwangi21:
Hi @eos87, found the bug and fixed it. Should be OK now.

server/liveblog/posts/utils.py
```diff
@@ -151,3 +151,42 @@ def get_main_item(post):
         )
 
     return main_item
+
+
+def check_content_diff(updates, original):
```
eos87, line 156: we should cover this one with tests please.

eos87 (approved):
I tested it and it works well. There is only one thing left which I added a comment for it. 
Well done!

client/app/scripts/liveblog-edit/posts.service.ts
```diff
@@ -359,9 +389,49 @@ const postsService = (api, $q, _userList, session) => {
...
+                        id: 'root',
+                        refs: [{ idRef: 'main' }],
+                        role: 'grpRole:NEP',
+                    }, {
+                        id: 'main',
+                        refs: [],
+                        role: 'grpRole:Main',
+                    },
+                ],
+            });
```
eos87, line 431: I see that have a soft delete at the moment but I noticed that you're doing a hard delete to the items and leaving a post without items referenced. We should either have soft delete for post and keep references or do hard delete to posts also. We can discuss this during our next dev meeting.

BrianMwangi21: Sure. No problem.

server/liveblog/posts/utils.py
BrianMwangi21: I've added tests for this in `server/liveblog/posts/tests/posts_test.py`

eos87 (approved):
Tested it and it works great!

PR #1362, opened 2024-03-07
[LBSD-2767] - Create the static UI to render the polls in Default SEO Theme

BrianMwangi21:
### Purpose

This PR implements the user interface for rendering polls in a static manner in the Default SEO theme. The goal is to provide a visual representations of polls, when both opened and closed, to enhance user engagement and interaction

### What has changed

- Added `template-poll.html` to render the polls
- Added `template-poll-utils.html` with helper function to calculate poll votes and time left in order to change poll view from open to closed
- Modified the client endpoints to call the `poll_calculation` function in order to pass the vote percentages in the item sent to the template for rendering when poll is closed

### Steps to test

3. Click "Live" to see the Live Preview. You will see the poll UI when open.
4. To view the poll UI when closed, wait for the poll time to elapse and it will change automatically to show the poll summary.

#### When Poll is Open
image

#### When Poll is Closed
image

Resolves: [LBSD-2767](https://sofab.atlassian.net/browse/LBSD-2767)


server/liveblog/blogs/blog.py
```diff
@@ -90,6 +91,11 @@ def fix_item_markup_if_needed(self, doc):
         if doc.get("type") == "text":
             original_text = doc["item"].get("text")
             doc["item"]["text"] = self.check_html_markup(original_text)
+        elif doc.get("type") == "poll":
+            doc["item"]["poll_body"]["active_until"] = doc["item"]["poll_body"][
+                "active_until"
+            ].isoformat()
```
eos87, line 97: As we spoke earlier today, please add some "safe guard" for potential `None` values.

server/liveblog/themes/themes_assets/default/templates/template-poll-utils.html
```diff
@@ -0,0 +1,30 @@
+<script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.29.1/moment.min.js"></script>
```
eos87, line 1: We already have `moment.js` as dependency in the theme. Why do you need it again like this here?

server/liveblog/themes/themes_assets/default/misc/filters.js
```diff
@@ -100,6 +100,15 @@ const decodeUri = (url) => {
...
 
+/**
+ * Converts a JavaScript object to a JSON string.
+ * @param  {Object} obj The object to be converted to JSON.
+ * @return {string} JSON string representation of the object.
+ */
+const tojson = (obj) => {
+  return JSON.stringify(obj);
+}
+
```
eos87, line 111: Filters need a counter part in the Python side, otherwise the embed generation will fail. See here

https://github.com/liveblog/liveblog/blob/master/server/liveblog/themes/template/filters.py
https://github.com/liveblog/liveblog/blob/master/server/liveblog/themes/themes.py#L291-L296

BrianMwangi21: I didn't add a counterpart filter on the Python side since Jinja already has an inbuilt filter of the same name : https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.tojson

eos87: Got it, thanks for clarifying.

eos87 (approved):

PR #1361, opened 2024-02-29
[LBSD-2765] - Implement the backend logic to display polls in the readers' side

BrianMwangi21:
### Purpose

This PR implements endpoints on the clients side to get polls within posts and allow readers to vote on them. 

### What has changed

- **Updated endpoint for getting posts:** The existing endpoint for fetching blogs posts on the clients side has been updated to  attach the poll data.

- **New Endpoint for Voting:** A patch endpoint has been introduced to enable readers to cast their votes on polls. This endpoint handles vote submission and updates the poll results.

### Steps to test

- Endpoints can be tested using a third party app or service like Postman. Also curl can be used to test the endpoints.

##### Get Client Posts
```
curl --location --request GET 'http://127.0.0.1:5000/api/v2/client_blogs/<blog_id>/posts'
```
##### Get Client Polls
```
curl --location --request GET 'http://127.0.0.1:5000/api/client_polls'
```
##### Get Client Poll
```
curl --location --request GET 'http://127.0.0.1:5000/api/client_polls/<poll_id>'
```
##### Patch Client Poll
```
curl --location --request PATCH 'http://127.0.0.1:5000/api/client_polls/<poll_id>' \
--header 'If-Match: <etag>' \
--header 'Content-Type: application/json' \
--data-raw '{
    "poll_body": {
        "answers": [
            {
                "option": "Yes",
                "votes": 750
            },
            {
                "option": "No",
                "votes": 450
            }
        ]
    }
}'
```

Resolves: [LBSD-2765](https://sofab.atlassian.net/browse/LBSD-2765)

[LBSD-2765]: https://sofab.atlassian.net/browse/LBSD-2765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1363, opened 2024-03-12
[LBSD-2768] - Connect the new readers' polls UI from Default SEO theme with the backend

BrianMwangi21:
### Purpose

This PR aims to integrate the UI components for the readers interface, developed in [LBSD-2767](https://github.com/liveblog/liveblog/pull/1362), with the backend endpoints implemented in [LBSD-2765](https://github.com/liveblog/liveblog/pull/1361).The goal is to enable readers to vote on the polls attached to blogs.

### What has changed

- Wired the frontend UI components for polls with the backend endpoints to allow poll voting.
- Added mechanism to prevent readers from voting more than once using the approach from the [research in LBSD-2766](https://sofab.atlassian.net/browse/LBSD-2766).

### Steps to test

3. Click "Live" to see the Live Preview. You will see the poll UI when open.
4. Select an option for the poll and click vote. This will trigger an event to place the vote and increment the total votes count. The selection will also be saved in storage for persistence.

#### When Poll is Open
image

#### When Reader Votes
image
_Note: Since the selection persists, even when you refresh, you shouldn't be able to vote again and your selection will be highlighted for a visual cue._

Resolves: [LBSD-2768](https://sofab.atlassian.net/browse/LBSD-2768)

[LBSD-2767]: https://sofab.atlassian.net/browse/LBSD-2767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LBSD-2765]: https://sofab.atlassian.net/browse/LBSD-2765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LBSD-2768]: https://sofab.atlassian.net/browse/LBSD-2768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eos87:
It seems that `decipherinc/angular-history` got removed from GitHub who knows why. That's making the build fail as it is a dependency of `superdesk-client-core`. I noticed that the SD guys have fixed it by putting the lib in the project so I will do the same with our fork https://github.com/superdesk/superdesk-client-core/pull/4454.

eos87 (changes requested):
I left one minor comment about typos. Also when I tested it I noticed some unexpected issues: 

- I voted in my poll but the count remained in 0, see first screenshot
- After voting, the options remain changeable as if I would be able to change my vote, I think this should not happen, once the user voted they should see the poll in the "results" mode

image

- After switching the order from Editorial to any other order, the vote button is available again. I was actually able to vote again. If I keep changing the order I can vote as many times as I want.

image

- Simple Theme is rendering the poll but broken. We should probably fix that in a separate PR once the default theme is fully adjusted.

eos87 (approved):
Tested it and it works great! Code looks good 🚀

PR #1367, opened 2024-03-20
[LBSD-2733] - Refactor `publish_blog_embed_to_s3` task

eos87:
### Purpose
This the first of two PRs to solve [LBSD-2733]. The second PR will be about giving feedback to client about the failure of generating the embed with the selected theme.

There were a couple of problems that have been fixed in this refactoring. Sometimes the `_publish_blog_embed_on_s3` function was provided without a theme (`None` value) which led to an error while generating the embed. While working on this refactoring I figured out that the theme parameter was really unnecessary therefore it has been removed. Now, to generate the embed, the function would use either the blog's theme or the output channel theme if provided.

The other issue that has been solved is that the `public_url` value of the blog was not being saved after the initial embed generation was processed. If the initial generation failed, no `public_url` was saved and subsequently it was not updated given a wrong logic. Now when the embed is generated and there is no `public_url` assigned, the blog will be updated anytime a new url is generated as part of the embed processing. This partly solves LBSD-2733 as when there is no `public_url` the Live button would not show up. 

### What has changed

- `publish_blog_embed_to_s3` task has been refactored for easier reading and maintainability
- introduced a couple of utility functions to try to declutter the code in this function which was a bit messy
- added some additional tests for such functions
- removed the theme parameter as it was not used (or at least I think it was not ;)) 

### Steps to test

- checkout the branch 
- run the services
- check that the embed generation works without any issues both with the regular blog embed and the output channels

Partly resolves: [LBSD-2733]

[LBSD-2733]: https://sofab.atlassian.net/browse/LBSD-2733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LBSD-2733]: https://sofab.atlassian.net/browse/LBSD-2733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/liveblog/blogs/utils.py
+++ b/server/liveblog/blogs/utils.py
@@ -153,3 +153,20 @@ def can_delete_blog(blog):
         return False
 
     return True
+
+
+def build_blog_public_url(app, blog_id, theme, output_id=None):
+    """
+    Creates the public url for a given blog. If `output_id` is provided
+    it will compile the url using the given theme
+    """
+    server_url = app.config["SERVER_NAME"]
+    protocol = app.config["EMBED_PROTOCOL"]
+
+    theme_str = output_str = ""
+
+    if output_id:
+        output_str = f"/{output_id}"
+        theme_str = f"/theme/{theme}" if theme else ""
+
+    return f"{protocol}{server_url}/embed/{blog_id}{output_str}{theme_str}"
--- a/server/liveblog/themes/themes.py
+++ b/server/liveblog/themes/themes.py
@@ -570,34 +570,59 @@ class ThemesService(BaseService):
         return response
 
     def publish_related_blogs(self, theme):
-        from liveblog.blogs.tasks import publish_blog_embed_on_s3
+        """
+        Publishes related blogs and outputs based on a given theme.
 
-        blogs = get_resource_service("blogs").find(
-            {"blog_preferences.theme": theme.get("name")}
-        )
-        outputs = get_resource_service("outputs").find({"theme": theme.get("name")})
-        countdown = 1
-        step = STEPS.get("default")
-        if theme.get("seoTheme"):
-            step = STEPS.get("seoTheme")
-        if theme.get("ampTheme"):
-            step = STEPS.get("ampTheme")
+        This function retrieves blogs and outputs that match the specified theme and schedules them for publishing.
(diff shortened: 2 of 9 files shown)


server/liveblog/blogs/utils.py
```diff
@@ -153,3 +153,20 @@ def can_delete_blog(blog):
         return False
 
     return True
+
+
+def build_blog_public_url(app, blog_id, theme, output_id=None):
```
BrianMwangi21, line 158:
```suggestion
def build_blog_public_url(app, blog_id, theme=None, output_id=None):
```

server/liveblog/themes/themes.py
BrianMwangi21, line 1: A suggested refactor for the `publish_related_blogs` function since the loop for blogs and output almost do the same thing : 

```
    def publish_related_blogs(self, theme):
        blogs = get_resource_service("blogs").find(
            {"blog_preferences.theme": theme.get("name")}
        )
        outputs = get_resource_service("outputs").find({"theme": theme.get("name")})
        step = STEPS.get("default")
        if theme.get("seoTheme"):
            step = STEPS.get("seoTheme")
        if theme.get("ampTheme"):
            step = STEPS.get("ampTheme")

        self._schedule_publish_related_blogs(blogs, step, is_blog=True)
        self._schedule_publish_related_blogs(outputs, step, is_blog=False)

        return blogs

    def _schedule_publish_related_blogs(self, items, step, is_blog=True):
        from liveblog.blogs.tasks import publish_blog_embed_on_s3
        countdown = 1
        for item in items:
            args = [item.get("_id")] if is_blog else [item.get("blog")]
            kwargs = {"output": item} if not is_blog else {}
            publish_blog_embed_on_s3.apply_async(args=args, kwargs=kwargs, countdown=countdown)
            countdown += step
```

BrianMwangi21 (commented):
Tested locally and works. Just left some suggestions.

server/liveblog/themes/themes.py
eos87: done

BrianMwangi21 (approved):
LGTM 👍

PR #1370, opened 2024-03-27
[LBSD-2733] - Completes fix for live button missing issue

eos87:
### Purpose

This is the 3rd and last PR to address the "Live button missing" issue [LBSD-2733] (other PRs #1367, #1369). Here we are trying to provide some useful feedback to the user in case the generation of the embed fails. Now instead of the button/link just going missing, the user would be informed about the error, a potential workaround will be suggested (try with a different theme) and to get in contact with out support team for further investigation about the issue. 

### What has changed

- Added a new web socket listener to receive embed generation errors. Once an error is received, the user is informed about through our UI notifications messages 
- The message informing the user about the error is only shown every 3 hours to avoid annoying the user too much as the error can keep happening multiple times every time the embed is updated
- Additional traceback information is logged (into syslog and sentry)

### Steps to test

The easiest way to test this would be to mock an exception in the embed generation function. Just throw the exception after line 47 in the code below, restart your services and then try to create a new post in your blog. Observe that the new message is showing in the UI.

https://github.com/liveblog/liveblog/blob/987a234d2da572a5775c1d8a68897ec483ebf75a/server/liveblog/blogs/tasks.py#L46-L57

### Screenshots
The new message looks like this:

|image|
|-|

Resolves: [LBSD-2733]

[LBSD-2733]: https://sofab.atlassian.net/browse/LBSD-2733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LBSD-2733]: https://sofab.atlassian.net/browse/LBSD-2733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


client/app/scripts/liveblog-edit/blog.service.js
```diff
@@ -61,10 +67,50 @@ export default function blogService(api, $q, $rootScope, config) {
         return deferred.promise;
     }
 
+    /**
+     * This function is triggered when a embed generation error is notified via web socket message.
+     * It makes sure that the notification is shown only once every 3 hours approximately by storing
+     * a key in local storage and checking if should show the error again or not. The intention
```
BrianMwangi21, line 73:
```suggestion
     * a key in local storage and checking if it should show the error again or not. The intention
```

BrianMwangi21 (commented):
Tested locally and it works.

PR #1372, opened 2024-04-09
[LBSD-2772] Fix author on syndicated blog not showing up properly

eos87:
### Purpose
To fix the issue that shows the wrong author on syndicated posts after these have been edited in the producer's side. 

### What has changed
- Adjustments made to the mechanism that sends the post to the consumer in order to attach the right creator's data
- Code refactor of syndication webhook code to make it easier to understand and maintain 
- Refactor of themes' variables name to give them a better meaning (from `item` to `post` in `template-post.html`)
- Adjusted themes to show the right author taking information from the syndicated_creator attribute instead of the first item of the post

### Steps to test

In order to test this PR you will need to set up Content Syndication in your local instance. You will first need to register a "Consumer" and then a "Producer" that would point out into your own local instance. Please read the wiki document about [Content Syndication](https://sofab.atlassian.net/wiki/x/hIFdPwI) to get a better understanding of it.

- Checkout the branch
- Set up Content Syndication
- Create a Producer blog (enable syndication in blog settings)
- Create a Consumer blog and attach the Producer blog (you can enable automatic publish)
- Create a post in Producer, see it ingested from the consumer 
- Then edit the same post with a different user. Observe that the author should remain the same both in Editor's UI and in the embed side

For the review, I would recommend you to hide the whitespace so there is less noise in the PR: 
| image |
| - |

Please let me know if you have any issues while trying to test it. 

Resolves: [LBSD-2772]

Thanks for checking!

[LBSD-2772]: https://sofab.atlassian.net/browse/LBSD-2772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


client/app/scripts/liveblog-edit/directives/posts-list.js
```diff
@@ -160,40 +161,66 @@ export default function lbPostsList(postsService, notify, $timeout, PagesManager
...
+            return false;
+
+        // Notify for scheduled post
+        notifyScheduledPostPublished(eventParams);
+
+        // Only update if posts belong to the same blog
+        updateIfPostsBelongToSameBlog(posts, $scope, listInstance, eventParams);
+    };
+
+    // check if it's neither the main timeline nor a comments panel with the first post syndicated
```
BrianMwangi21, line 186:
```suggestion
    /**
     * Determines if the element should be updated based on: 
     * 1. If it not the main timeline
     * 2. It is not a comments panel
     * 3. If the first post is syndicated
     */
```

server/liveblog/syndication/utils.py
```diff
@@ -139,48 +134,64 @@ def extract_post_items_data(original_doc):
...
+    users_service = get_resource_service("users")
+    original_creator = users_service.find_one(req=None, _id=doc["original_creator"])
+
+    if not original_creator:
+        return None
+
+    creator_data = {}
+    for key in needed_fields:
+        if key in original_creator:
+            creator_data[key] = original_creator.get(key)
```
BrianMwangi21, line 146: Suggestion to use comprehension: 

`creator_data = {key: original_creator[key] for key in needed_fields if key in original_creator}`

server/liveblog/syndication/syndication.py
```diff
@@ -236,95 +236,129 @@ class SyndicationIn(Resource):
...
+            return api_error(
+                "Updates should not be sent for a blog which is not open", 409
+            )
+
+        # ensure necessary data is in the request
+        self.request_data = data = request.get_json(silent=True) or {}
+        if not all(key in data for key in ("items", "post")):
+            return api_error("Bad Request", 400)
+
+        return super(SyndicationWebhook, self).dispatch_request(*args, **kwargs)
```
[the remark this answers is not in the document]
eos87: No reason really, just an old python 2 habit. I will fix it

server/liveblog/syndication/utils.py
eos87: I'm not a huge fan of `comprehension` because I think they do not contribute much in terms of readability but, your suggestion makes sense from the "performance" point of view. Nevertheless, given that the amount of data we process here is really small, I'd argue that the performance differences would be negligible therefore I would be more inclined into readability over performance.

What do you think?

BrianMwangi21: I think then it's fine then if the amount of data processed here is small. That's ok with me

BrianMwangi21 (approved):

server/liveblog/syndication/syndication.py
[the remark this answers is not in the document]
eos87: fixed.

client/app/scripts/liveblog-edit/directives/posts-list.js
eos87: fixed

client/app/scripts/liveblog-edit/directives/posts-list.js
```diff
@@ -160,40 +161,66 @@ export default function lbPostsList(postsService, notify, $timeout, PagesManager
...
+    const isNotMainTimelineNorCommentsWithFirstPostSyndicated = ($element, $scope, eventParams) => {
```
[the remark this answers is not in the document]
eos87: fixed

PR #1375, opened 2024-04-17
[LBSD-2782] - Implement new backend structure to store Instance Settings

BrianMwangi21:
### Purpose

This PR introduces a new database collection to store instance settings in JSON format. The purpose of these changes is to ensure that instance configurations are both flexible and robust, allowing settings to be defined and adjusted during deployment and throughout the lifecycle of the instance. The new settings mechanism includes validation to prevent configurations that could disrupt application functionality.

### What has changed

- New DB Collection: Implemented a collection in the database to store instance settings as JSON. This allows dynamic and flexible configuration management.
- Schema Validation: Added schema validation logic to ensure the integrity of the JSON data, protecting the application from potential misconfiguration.
- Default Configuration Fixture: Introduced a fixture file containing default instance settings. This fixture ensures that necessary settings are established during initial deployment.
- Conditional Update Logic: Developed logic to compare existing instance settings against the default provided in the fixture. This update process preserves existing settings by only adding new keys that are missing from the current configuration.

### Steps to test

1. Check out to the branch
2. Send a POST request to add new instance settings: 
```
curl --location --request POST 'http://127.0.0.1:5000/api/instance_settings' \
--header 'Authorization: <basic_auth_token>' \
--header 'Content-Type: application/json' \
--data-raw '{
    "settings": {
       "solo": {
            "features": {
                "syndication": true,
                "marketplace": false
            },
            "limits": {
                "blogs": 1,
                "themes": 1,
                "blog_members": 1
            }
        },
        "team": {
            "features": {
                "syndication": true,
                "marketplace": true
            },
            "limits": {
                "blogs": 3,
                "themes": 6,
                "blog_members": 5
            }
        },
        "network": 1
    }
}'
```
3. Send a GET request to view instance settings:
```
curl --location --request GET 'http://127.0.0.1:5000/api/instance_settings' \
--header 'Authorization: <basic_auth_token>' 
```
4. Send another POST request to updated existing settings giving precedence to existing config values:
```
curl --location --request POST 'http://127.0.0.1:5000/api/instance_settings' \
--header 'Authorization: <basic_auth_token>' \
--header 'Content-Type: application/json' \
--data-raw '{
    "settings": {
        "one-time-event": {
            "features": {
                "syndication": false,
                "marketplace": false
            },
            "limits": {
                "blogs": 1,
                "themes": 1,
                "blog_members": 2
            }
        },
        "uni": {
            "features": {
                "syndication": true,
                "marketplace": false
            },
            "limits": {
                "blogs": 1,
                "themes": 2,
                "blog_members": 3
            }
        },
        "solo": {
            "features": {
                "syndication": true,
                "marketplace": false
            },
            "limits": {
                "blogs": 1,
                "themes": 14,
                "blog_members": 1
            }
        },
        "team": {
            "features": {
                "syndication": true,
                "marketplace": true
            },
            "limits": {
                "blogs": 3,
                "themes": 6,
                "blog_members": 5
            }
        },
        "network": 1
    }
}'
```
4. Send another GET request to view instance settings. You will note that both `uni` and `one-time-event` have been added to the existing config. Also, you will note that even though the new request has set `solo.limits.themes` as 14, the previous value of 1 is retained.
5. To test validation, send a POST request with either a missing key or a non-integer value such as the following: 
```
curl --location --request POST 'http://127.0.0.1:5000/api/instance_settings' \
--header 'Authorization: <basic_auth_token>' \
--header 'Content-Type: application/json' \
--data-raw '{
   "settings": {
        "one-time-event": {
            "features": {
                "syndication": false,
                "marketplace": "false"
            },
            "limits": {
                "blogs": 1,
                "themes": 0
            }
        }
    }
}'
```
You should get the following error response listing all areas with an error: 
```
{
    "_error": {
        "code": 400,
        "message": "Validation errors for config occurred: The 'marketplace' in 'features' for 'one-time-event' must be a <class 'bool'>., The 'themes' in 'limits' for 'one-time-event' must be at least 1., Missing 'blog_members' in 'limits' for 'one-time-event'."
    },
    "_status": "ERR"
}
```

Resolves: [LBSD-2782](https://sofab.atlassian.net/browse/LBSD-2782)

[LBSD-2782]: https://sofab.atlassian.net/browse/LBSD-2782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/liveblog/validator.py
+++ b/server/liveblog/validator.py
@@ -89,3 +89,36 @@ class LiveblogValidator(SuperdeskValidator):
                             response.status_code
                         ),
                     )
+
+    def _validate_settings(self, settings):
+        errors = []
+        required_schema = {
+            "max_active_blogs": {"type": "integer", "min": 1, "required": True},
+            "max_blog_members": {"type": "integer", "min": 1, "required": True},
+            "max_themes": {"type": "integer", "min": 1, "required": True},
+        }
+
+        for key, value in settings.items():
+            if key == "network":
+                # Special case for 'network' plan
+                if not isinstance(value, int) or value < 1:
+                    errors.append(
+                        f"The value for '{key}' must be an integer and at least 1."
+                    )
+            else:
+                # General case for other plans
+                if not isinstance(value, dict):
+                    errors.append(f"The value for '{key}' must be a dictionary.")
+                else:
+                    for field, attrs in required_schema.items():
+                        if field not in value:
+                            errors.append(f"Missing '{field}' in settings for '{key}'.")
+                        elif (
+                            not isinstance(value[field], int)
+                            or value[field] < attrs["min"]
+                        ):
+                            errors.append(
+                                f"The '{field}' for '{key}' must be an integer and at least {attrs['min']}."
+                            )
+
+        return errors
--- a/server/liveblog/validator.py
+++ b/server/liveblog/validator.py
@@ -105,35 +105,27 @@ class LiveblogValidator(SuperdeskValidator):
         }
 
         for key, config in settings.items():
-            if key == "network":
-                # Special case for 'network' plan
-                if not isinstance(config, int) or config < 1:
-                    errors.append(
-                        f"The value for '{key}' must be an integer and at least 1."
-                    )
-            else:
-                # General case for other plans
-                for section, section_config in required_schema.items():
-                    if section not in config:
-                        errors.append(f"Missing '{section}' in settings for '{key}'.")
-                    else:
-                        for field, field_attrs in section_config.items():
-                            if field not in config[section]:
+            for section, section_config in required_schema.items():
+                if section not in config:
+                    errors.append(f"Missing '{section}' in settings for '{key}'.")
+                else:
+                    for field, field_attrs in section_config.items():
+                        if field not in config[section]:
+                            errors.append(
+                                f"Missing '{field}' in '{section}' for '{key}'."
+                            )
+                        else:
+                            field_value = config[section][field]
+                            if not isinstance(field_value, field_attrs["type"]):
                                 errors.append(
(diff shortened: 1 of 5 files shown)


server/liveblog/validator.py
```diff
@@ -89,3 +89,51 @@ def _validate_httpsurl(self, httpsurl, field, value):
...
+        }
+
+        for key, config in settings.items():
+            if key == "network":
+                # Special case for 'network' plan
+                if not isinstance(config, int) or config < 1:
+                    errors.append(
+                        f"The value for '{key}' must be an integer and at least 1."
+                    )
+            else:
```
eos87, line 114: Maybe we can remove this one out and we can take care of it in the features service that I'm currently working on. Thoughts?

eos87 (approved):
It looks good to me. I just left one comment, please take a look and let me know what you think.

server/liveblog/validator.py
BrianMwangi21: Hmm, I think we still need it here since the 'network' value is just an int and not a dict like the other plans so the validation will fail if we remove this part and leave the rest logic that checks the dicts. Unless of course we change network to also be a dict with "features" and "limits". 

What do you think ?

eos87: yeah but my thinking was more about getting rid of network from the settings and only returning "true" when checking limitations if the current subscription level is network, even without checking the settings. That way we could avoid some complexity here.

BrianMwangi21: Oh, so essentially the config will have all other plans except network then on the features service it handles that case there. Makes sense

PR #1379, opened 2024-04-26
[LBSD-2785] - Implement features service to check for subscription limits

eos87:
Implements a new service, both backend and frontend, that will be used to check if certain features are enabled or to check the limitations of the current user's subscription. 

### What has changed
- Added a new python service class that fetches the settings from database and does a simple caching logic to avoid hitting database too often
- Implemented a new simple api endpoint to retrieve instance settings based on current user's subscription
- Added a new Typescript class and a new Angular Service based on this class that consumes the new endpoint and implement a couple of helper methods to check if features are enabled and the limitation to restrict the behaviour of the UI
- Adjusted the app's code to use the new features services to check if features are enabled or some limits have been reached 

### Steps to test
- Checkout the branch and run the project
- Switch between the different subscription plans to see if the limitations are all in place
- Test the app in general to see if the there are any issues that might have been introduced buy these changes

Resolves: [LBSD-2785]

Thanks for checking!

[LBSD-2785]: https://sofab.atlassian.net/browse/LBSD-2785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BrianMwangi21:
Testing it locally and found an error when I was in another plan other than network and tried to create an extra blog. I also did not see the upgrade modal 
image

client/webpack.config.js
```diff
@@ -244,24 +242,8 @@ const configApp = (grunt) => ({
...
-    excludedTheme: 'angular',
-    assignableUsers: {
-        solo: 2,
-        team: 4
-    },
-    subscriptionLevel: process.env.SUBSCRIPTION_LEVEL || '',
-    blogCreationRestrictions: {
-        solo: 1,
-        team: 3
-    },
```
BrianMwangi21, line 264: I'm guessing this section has been removed since it was redundant with similar options in `options/template.js` right ?

eos87: Indeed!

eos87:
> Testing it locally and found an error when I was in another plan other than network and tried to create an extra blog. I also did not see the upgrade modal image
> 
> Might be related to the features service initialize not working as I'm getting a CORS error. Weird cause other requests are going through image

Hm, this is weird. I will take a look and get back to you later today.

client/app/scripts/liveblog-features.service.ts
BrianMwangi21, line 1: 👍

server/liveblog/instance_settings/features_service.py
BrianMwangi21, line 1: 👍

BrianMwangi21 (commented):
Code looks good. Just testing it out locally that's left

BrianMwangi21 (approved):

PR #1381, opened 2024-05-03
[LBSD-2786] - Complete Instance Settings UI

eos87:
### Purpose
Completes the implementation of the UI to manage Instance Settings. This UI is restricted only to users that have the `is_support` attribute active. If the attribute is not active, then the UI is not visible.

### What has changed
- The resource schema has been enhanced
- Validator method has been removed in favour of using Cerberus (Python Eve) dependency to run the validations against the resource's schema 
- Improved service class to make the instance settings act as a singleton
- Implemented a few edge case when settings are not present
- Improved frontend controller and fixed linter issues
- Removed `Instance Admin` role in favour of using `is_support` attribute instead
- Render instance settings UI based on user's `is_support` attribute
- Added validation in the backend side so user cannot modify the instance settings if it doesn't have the `is_support` enabled
- Remove the `merge_config` method to keep it simple for now. In the future we can analyse if a more complex approach is needed
- Adjusted couple of places with new features service (I've missed them in previous PR)
- Adjusted tests

### Steps to test
- Checkout the branch
- Run the services
- Go to instance settings UI. Notice that if user is not support, the menu entry will not render. Otherwise it should
- Try to save the setting and play with validations. The settings structure should look like this below. You can try removing "features" for example, and the system should catch that.

```json
{
    "settings": {
        "solo": {
            "features": {
                "syndication": true,
                "marketplace": false,
                "theme_styles": false,
                "custom_themes": false,
                "change_blog_theme": false,
                "freetypes_manager": false,
                "archived_blogs_available": false
            },
            "limits": {
                "blogs": 1,
                "themes": 1,
                "blog_members": 1,
                "custom_themes": 0
            }
        },
        "team": {
            "features": {
                "syndication": true,
                "marketplace": true,
                "theme_styles": true,
                "custom_themes": true,
                "change_blog_theme": true,
                "freetypes_manager": true,
                "archived_blogs_available": true
            },
            "limits": {
                "blogs": 3,
                "themes": 6,
                "blog_members": 5,
                "custom_themes": 5
            }
        }
    }
}
```

Thanks for reviewing!

Resolves: [LBSD-2786]

[LBSD-2786]: https://sofab.atlassian.net/browse/LBSD-2786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BrianMwangi21 (approved):
I've tested it locally and it works. LGTM! 👍

eos87:
I will merge this although there will be another PR just to adjust the plans' names and features/limits (Liveblog Go and Uni)

PR #1383, opened 2024-05-20
[LBSD-2807] - Critical miscalculation of Poll voting results 

BrianMwangi21:
### Purpose

This PR fixes a bug which was resulting in the wrong vote being placed due to mismatch of order of answers between rendering the poll and subsequent voting.

### What has changed
- Use of option value instead of index value to confirm vote selection and add votes to a specific option.

### Steps to test
1. Log in to the Editor admin.
2. Create a new poll with 3 options.
3. Open the Live preview of the blog in 3 different browsers.
4. In the first browser, vote for the first option. In the second browser, vote for the last option. 
5. Before voting on the third browser, open the live preview again in a fourth browser. Observe that the two options you have already voted for are at the top. This is because the poll calculations in the background change the options ordering.
6. Observe that on the initial third browser, the original render remains unchanged. Proceed to vote for the second option.
7. Observe that the votes will be distributed accordingly. 

Resolves: [LBSD-2807](https://sofab.atlassian.net/browse/LBSD-2807)


eos87 (approved):
I didn't test this locally but code wise looks good.

PR #1384, opened 2024-05-20
[LBSD-2809] - Error is thrown when trying to edit a syndicated poll

BrianMwangi21:
### Purpose

This PR fixes a bug which was causing updates on Polls to fail on the consumer side after syndication. This issue was caused because the consumer poll was being created on the items service instead of the poll service.  This PR also fixes a small bug of showing the ingestion indicator on the Producer side.

### What has changed

- Changed the `create_syndicated_blog_post` function to accommodate for the polls to be created by the polls service and not the items service before creating the post for the consumer. 
- Show ingestion indicator on Consumer side only.

### Steps to test

1. Log in to the Editor admin as a producer and create a poll.
2. Log in to the Editor admin as a consumer and ingest the new post from the producer.
3. Try editing the poll from the timeline i.e change the duration of the poll. Observe that it updates correctly. 

Resolves: [LBSD-2709](https://sofab.atlassian.net/browse/LBSD-2809)

[LBSD-2709]: https://sofab.atlassian.net/browse/LBSD-2709?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/liveblog/syndication/utils.py
```diff
@@ -318,14 +319,25 @@ def create_syndicated_blog_post(producer_post, items, in_syndication):
...
+        item_service = get_resource_service("blog_items")
+        item_ids = item_service.post(post_items)
+        item_refs.extend({"residRef": str(item_id)} for item_id in item_ids)
+
+    if post_polls:
+        poll_service = get_resource_service("blog_polls")
+        poll_ids = poll_service.post(post_polls)
+        item_refs.extend(
+            {"residRef": str(poll_id), "location": "polls"} for poll_id in poll_ids
+        )
```
eos87, line 340: This looks very redundant to me. Is there a way we can make this dynamic somehow? We will have more complex item types like polls in the future.

server/liveblog/polls/polls.py
```diff
@@ -157,3 +159,23 @@ def poll_calculations(poll):
...
+    resource_methods = ["GET"]
+    privileges = {"GET": "posts"}
+
+
+class BlogPollsService(BaseService):
+    def get(self, req, lookup):
+        if lookup.get("blog_id"):
+            lookup["blog"] = ObjectId(lookup["blog_id"])
+            del lookup["blog_id"]
+        return super().get(req, lookup)
```
eos87, line 181: What is really the reason for the extra Resource and Service here?

eos87 (changes requested):

eos87 (approved):
Looks good

PR #1387, opened 2024-05-30
Release/v3.85.1

eos87:
### Purpose
Release Liveblog `v3.85.1`

### What has changed
- Fixes instagram embeds rendering 
- Lock `amphtml-validator` for compatibility issues
- Fixes `angular-history` repo which is no longer available


BrianMwangi21 (approved):
LGTM. Tested out locally too and it works.

PR #1391, opened 2024-06-13
Change `newest_first` and `oldest_first` ordering attribute

eos87:
### Purpose
There are some cases when the users create a post but do not publish it right away (like scheduled posts). Such posts, could end up not being rendered in the right position when using Newest First or Oldest First tabs in the readers side as it was using `_created` attribute which is assigned at the time the entry is added into the database. To avoid such issues, we need to use `published_date` attribute instead, as this is the one that stores the real date and time when the post is available to the readers

### What has changed
The ordering attribute for the embed generation has been changed from `_created` to `published_date`

### How to test
- Checkout the branch and run the project
- Create a blog, set default ordering to "Newest First" in Theme Settings section
- Create multiple posts. Add a scheduled one that would end up in the middle while you create a publish other posts
- Observe that the scheduled post renders in the right position after reloading the embed side

Resolves: [LBSD-2824]

[LBSD-2824]: https://sofab.atlassian.net/browse/LBSD-2824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -36,7 +36,7 @@ jobs:
         ports:
           - 6379:6379
       elasticsearch-1.7.2:
-        image: barnybug/elasticsearch:1.7.2
+        image: liveblog3/elasticsearch:1.7.2
         ports:
           - 9200:9200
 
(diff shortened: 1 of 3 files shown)


BrianMwangi21 (approved):
Tested locally and it works

brianFabric (approved):

PR #1404, opened 2024-07-26
[LBSD-2836] - Change latestUpdate to use current time on embed

BrianMwangi21:
### Purpose

This PR changes the implementation on the embed to set the latestUpdate variable as the time when the embed loads. Following this, all post before that time are fetched and rendered on the timeline.

### What has changed

- Changed `latestUpdate` to be the current time on loading embed
- Added an `initialRender` function which loads post before the current time and renders on the timeline after which new posts are fetched and rendered normally.

### Steps to test

1. Open the Editor.
2. Publish normal and schedule posts at different intervals. 
3. Observe on the live preview that the post shows as intended. 
4. To test if the latestUpdate is the current time of page loading, open the network tab on the developer tools and check the first API call to the `/posts` endpoint. The payload query should look like this with the current time set in latestUpdate used in the range value : 
```
{"query":{"filtered":{"filter":{"and":[{"term":{"sticky":false}},{"term":{"post_status":"open"}},{"range":{"_updated":{"lte":"2024-07-26T08:04:51.318Z"}}}]}}},"sort":[{"order":{"order":"desc","missing":"_last","unmapped_type":"long"}}]}
```

Resolves: [LBSD-2836](https://sofab.atlassian.net/browse/LBSD-2836)


eos87 (approved):
This looks good to me and it makes sense.

PR #1407, opened 2024-08-09
[LBSD-2832] - Create Go pricing plan notification for bandwidth

BrianMwangi21:
### Purpose

This PR introduces an implementation to fetch bandwidth usage for Go pricing plan and notify the user on the client side. 

### What has changed

- Added a new `Bandwidth` resource and service to store the bandwidth of the client. 
- Added a new task to make the API call to fetch bandwidth usage from Cloudflare and scheduled it for midnight. 
- Added endpoint to fetch bandwidth usage on client side and notify the user when close to upper bandwidth limit

### Steps to test
1. Checkout to this branch and set needed env variables for the server, then run Liveblog. Ensure the subscription level is 'liveblog-go'.
```
SUBSCRIPTION_LEVEL='liveblog-go'
CLOUDFLARE_URL=''
CLOUDFLARE_AUTH=''
CLOUDFLARE_ZONE_TAG=''
```
2. Under the server directory, initialize data to create new bandwidth collection 
```
$ python3 manage.py app:initialize_data;
```
3. For easier testing, you can change the schedule of the task to run every two minutes under `settings.py`
```
    "bandwidth:update": {
        "task": "liveblog.bandwidth.tasks.fetch_bandwidth_usage",
        # "schedule": crontab(minute=0, hour=0),  # every midnight
        "schedule": crontab(minute="*/2"),  # every 2 minutes
    },

```
4. Observe the logs to see the task being called and running. 
5. After a while, log in on the client side and if the bandwidth used is greater than 75%, it should show a warning notification. You can reduce this value in `client/app/scripts/liveblog-bloglist/controllers/blog-list.js` for easier testing.

Resolves: [LBSD-2832](https://sofab.atlassian.net/browse/LBSD-2832)

[LBSD-2832]: https://sofab.atlassian.net/browse/LBSD-2832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


client/app/scripts/liveblog-bloglist/controllers/blog-list.js
```diff
@@ -489,4 +489,13 @@ export default function BlogListController(
...
     // Reload bloglist on membership's approval
     $scope.$on('blogs', fetchBlogs);
+
+    // Check for bandwidth
+    api.get('/bandwidth/current').then((bandwidthUsage) => {
+        const percentageUsed = parseFloat(bandwidthUsage.percentageUsed.toFixed(1));
+
+        if (percentageUsed >= 90) {
+            notify.error(`Warning: You have used ${percentageUsed}% of your allocated bandwidth.`);
+        }
```
eos87, line 499: I'd suggest improving this a little bit more. Like a modal dialog and show more information. Also telling the customer that we'll reach out to them soon. @aleksandarjelicic what do you think?

server/liveblog/bandwidth/bandwidth.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    Returns the bandwidth for the instance
+    """
+    current_bandwidth = get_resource_service(bandwidth_key).get_current_bandwidth()
+
+    response = {}
+    if current_bandwidth:
+        bandwidth_usage_bytes = current_bandwidth["bandwidthUsage"]
+        bandwidth_usage_gb = bandwidth_usage_bytes / (1024**3)
+        upper_limit_gb = 0.5  # TODO: This should be configurable, or set in feature limits of subscription
```
eos87, line 86: Yeah, this should come from instance settings. You could set `unlimited` or a number to specify the number of GBs.

server/liveblog/bandwidth/tasks.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@celery.task
+def fetch_bandwidth_usage():
+    """
+    Fetch bandwidth usage from Cloudflare API
+    """
+    logger.info("Fetching bandwidth usage from Cloudflare API")
+
+    if SUBSCRIPTION_LEVEL != SUBSCRIPTION_LEVEL_GO:
```
eos87, line 26: can we instead check the instance settings like "has bandwidth limit" instead? that way it wouldn't be hardcoded only to 'GO'

server/settings.py
```diff
@@ -152,6 +152,10 @@ def env(variable, fallback_value=None):
         "task": "liveblog.blogs.tasks.remove_deleted_blogs",
         "schedule": crontab(hour="*/1"),  # run every hour
     },
+    "bandwidth:update": {
+        "task": "liveblog.bandwidth.tasks.fetch_bandwidth_usage",
+        "schedule": crontab(minute=0, hour=0),  # every midnight
```
eos87, line 157: can we do this more often? If there is a busy liveblog instance, the limit might be consumed quickly and by the time we know they crossed the limit, their event could be over :)

client/app/scripts/liveblog-bloglist/controllers/blog-list.js
aleksandarjelicic: @BrianMwangi21 Where is this meassage showing now?
Modal dialog can be limiting, you have to close it to continue working, and it's hard to see it again.

BrianMwangi21: @aleksandarjelicic it's showing a small notification for now 
image

I was thinking maybe we can have an alert like this one: 

image

What do you think ?

aleksandarjelicic: @BrianMwangi21 second solution is good. You can put it in top black bar, or even above it.

server/settings.py
```diff
@@ -152,6 +152,10 @@ def env(variable, fallback_value=None):
         "task": "liveblog.blogs.tasks.remove_deleted_blogs",
         "schedule": crontab(hour="*/1"),  # run every hour
     },
+    "bandwidth:update": {
+        "task": "liveblog.bandwidth.tasks.fetch_bandwidth_usage",
+        "schedule": crontab(minute=0, hour="*/1"),  # every hours
+    },
```
eos87, line 158: Similarly here, this task should only be scheduled if the subscription level requires it. Since we cannot access the instance settings at this point (before scheduling it), we either add a condition inside the task itself that would check the setting or we schedule it only if the subscription level is "liveblog-go".

eos87 (approved):

PR #1410, opened 2024-09-03
[LBSD-2861] - "No. of blogs using theme: " is always set to zero 

BrianMwangi21:
### Purpose

This PR fixes the implementation of showing the number of blogs using a particular theme and shows the accurate number instead of 0.

### What has changed

- Fixed code implementation for getting the total blogs using a particular theme.

### Steps to test

1. Open the Editor
2. Create several blogs with several themes.
3. Open the side menu and head to the "Theme Manager". 
4. Observe that for each theme, the number of blogs using the theme is accurate.

### Screenshots

| Original  | New |
| ------------- | ------------- |
| image | image |

Resolves: [LBSD-2861](https://sofab.atlassian.net/browse/LBSD-2861)

### DIFF
--- a/client/app/scripts/liveblog-themes/module.js
+++ b/client/app/scripts/liveblog-themes/module.js
@@ -17,6 +17,7 @@ import listTpl from 'scripts/liveblog-themes/views/list.ng1';
         'upload',
         'privileges',
         'featuresService',
+        '$q',
     ];
     function LiveblogThemesController(
         _,
@@ -33,7 +34,8 @@ import listTpl from 'scripts/liveblog-themes/views/list.ng1';
         session,
         upload,
         privileges,
-        featuresService
+        featuresService,
+        $q
     ) {
         const self = this;
         /**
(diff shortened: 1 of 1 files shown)


client/app/scripts/liveblog-themes/module.js
```diff
@@ -115,53 +117,57 @@ import listTpl from 'scripts/liveblog-themes/views/list.ng1';
...
             // load only global preference for themes.
-            api.global_preferences.query({where: {key: 'theme'}}).then((globalPreferences) => {
-                self.globalTheme = _.find(globalPreferences._items, (item) => item.key === 'theme');
+            const globalPrefPromise = api.global_preferences.query({where: {key: 'theme'}})
+                .then((globalPreferences) => {
+                    self.globalTheme = _.find(globalPreferences._items, (item) => item.key === 'theme');
+                });
+
+            // get all blogs
+            const blogsPromise = api.blogs.query().then((data) => {
```
eos87, line 126: wouldn't this fetch like 25 results by default?

eos87 (commented):
I'm wondering if this should be better solved in the python side. Perhaps it will be simpler? What do you think?

client/app/scripts/liveblog-themes/module.js
BrianMwangi21: Yes. It fetches all the blogs once and then filters them for each theme in this section: 

```
themes.forEach((theme) => {
    const matchingBlogs = self.blogs.filter((blog) =>
        blog.blog_preferences
        && blog.blog_preferences.theme === theme.name);

    theme.blogs_count = matchingBlogs.length;
    theme.blogs = matchingBlogs;
```

The criteria we were using before wasn't working even if you tried it on postman and would always returned 0 for each theme. 

At the moment I'm trying to figure why that is and I suspect it is because on the `blog_schema` the mapping for `blog_preferences` is set to False so the `blog_preferences.theme` is not queried against search operations. So I'm looking into it as a separate issue considering `blog_preferences.theme` is used in multiple places.

PR #1412, opened 2024-09-10
[LBSD-2864] - Update post on poll  vote

BrianMwangi21:
### Purpose

This PR improves on the Poll feature where when a vote is placed, it also updates the `content_update_date` so that the embeds can capture the update and display the updates in votes on the timeline.

### What has changed

- Update the `content_update_date` on the parent post of a poll. 

### Steps to test

2. Open the Live Preview in multiple browsers. 
3. Place votes randomly and observe that the other clients show the updates of the total votes and the voting tally periodically

### Video
[polls-update.webm](https://github.com/user-attachments/assets/d9cdd882-f61b-4600-ad76-ec0d5d1e5e3f)

Resolves: [LBSD-2864](https://sofab.atlassian.net/browse/LBSD-2864)

[LBSD-2864]: https://sofab.atlassian.net/browse/LBSD-2864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/liveblog/client_modules/client_modules.py
+++ b/server/liveblog/client_modules/client_modules.py
@@ -472,9 +472,9 @@ def create_amp_comment():
     client_domain = data.get("__amp_source_origin")
     resp.headers["Access-Control-Allow-Origin"] = client_domain
     resp.headers["AMP-Access-Control-Allow-Source-Origin"] = client_domain
-    resp.headers["Access-Control-Expose-Headers"] = (
-        "AMP-Access-Control-Allow-Source-Origin"
-    )
+    resp.headers[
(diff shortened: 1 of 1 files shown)


eos87 (changes requested):
This would trigger the embed update (`update_post_blog_embed.delay(post)`) which could be a problem if there are thousands of votes, it would potentially jam our celery workers therefore resources.

eos87 (approved):
Looks good

PR #1413, opened 2024-09-30
[LBSD-2878] - Add "no-cache" for poll requests on embed

BrianMwangi21:
### Purpose

This PR adds "no-cache" to the Cache control attribute for poll requests so that they re-validate with the server to avoid the use of stale data. 

### What has changed

- Added "no-cache" attribute to poll requests on the embed

### Steps to test

2. Open the preview and place a vote.
3. Observe that the request to fetch a poll first to get the etag before patching has the "no-cache" attribute in the Cache Control option in the request headers

Resolves: [LBSD-2878](https://sofab.atlassian.net/browse/LBSD-2878)

[LBSD-2878]: https://sofab.atlassian.net/browse/LBSD-2878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/liveblog/themes/themes_assets/default/js/theme/helpers.js
```diff
@@ -47,6 +47,7 @@ function getJSON(url) {
     var xhr = new XMLHttpRequest();
 
     xhr.open('GET', url);
+    xhr.setRequestHeader("Cache-Control", "no-cache");
```
eos87, line 50: this will add no-cache to every request, isn't it? that would have a significant impact on our server side.

eos87 (changes requested):

server/liveblog/themes/themes_assets/default/js/theme/helpers.js
BrianMwangi21: I added this so that after voting, the results reflect immediately on the next cycle of checking for new posts and not after a few api calls as it is now because of the cache. 

Or do you think we can still proceed with it that way ?

eos87: I understand the reasoning but I think this `getJSON` function is used for different things not only the polls, right?

BrianMwangi21: It's used mainly when getting all posts. 

The other usecases are: 
1. In `ads-manager` during init.
2. Getting a single post and here it is for the new share functionality.

eos87 (approved):
Good to go after some minor changes requested

PR #1415, opened 2024-10-04
[LBSD-2876] - After switching to child theme and getting error, going back to Default theme is not possible

BrianMwangi21:
### Purpose

This PR adds a fallback mechanism to revert back to default theme if an embed generation fails on any other blog theme generation.

### What has changed

- Added fallback mechanism for blog embed generation to default theme.
- Updated Simple and AMP theme to use latest dev version of default theme.

### Steps to test

The easiest way to test this would be to mock an exception in the embed generation function. Just throw the exception after line 47 in the code below, restart your services and then update the blog theme. Observe the notification on the UI. Also observe that on refresh the theme is set as the Default theme.

https://github.com/liveblog/liveblog/blob/987a234d2da572a5775c1d8a68897ec483ebf75a/server/liveblog/blogs/tasks.py#L46-L57

### Screenshots of notification to user with fallback info

image 

Resolves: [LBSD-2876](https://sofab.atlassian.net/browse/LBSD-2876)

[LBSD-2876]: https://sofab.atlassian.net/browse/LBSD-2876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/app/scripts/liveblog-edit/blog.service.js
+++ b/client/app/scripts/liveblog-edit/blog.service.js
@@ -88,7 +88,7 @@ export default function blogService(api, $q, $rootScope, config, notify) {
             const errorMessage = `There has been an error while generating the embed for the
                 current blog. Please try with another theme and contact our Support Team
                 and provide this error: "${eventData.error}". The embed generation has fallen
-                back to the Default theme for the current blog.`;
+                back to the Default SEO theme for the current blog.`;
 
             notify.error(errorMessage, 30000);
             Storage.write(storeKey, '1', expiry);
--- a/client/app/scripts/liveblog-edit/blog.service.js
+++ b/client/app/scripts/liveblog-edit/blog.service.js
@@ -87,7 +87,8 @@ export default function blogService(api, $q, $rootScope, config, notify) {
 
             const errorMessage = `There has been an error while generating the embed for the
                 current blog. Please try with another theme and contact our Support Team
-                and provide this error: "${eventData.error}"`;
+                and provide this error: "${eventData.error}. The embed generation has fallen
+                back to the Default theme for the current blog."`;
 
             notify.error(errorMessage, 30000);
             Storage.write(storeKey, '1', expiry);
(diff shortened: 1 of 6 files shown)


server/liveblog/blogs/tasks.py
```diff
@@ -54,7 +55,33 @@ def publish_embed(blog_id, theme=None, output=None, api_host=None):
             f"Failed embed generation with theme `{theme}`. Error: {e}",
             exc_info=exc_info,
         )
-        return notify_about_embed_generation_error(str(e), blog_id, theme)
+        notify_about_embed_generation_error(str(e), blog_id, theme)
+
+        logger.info('generate_embed_fallback for blog "{}" started.'.format(blog_id))
+        try:
+            theme = "default"
```
eos87, line 62: If for some reason the theme failing is "default" theme, we would end up in an infinite loop of embed generation tasks. So we need to skip this if the theme is already default plus avoid telling the client that it will fallback to Default.

client/app/scripts/liveblog-edit/blog.service.js
```diff
@@ -87,7 +87,8 @@ export default function blogService(api, $q, $rootScope, config, notify) {
 
             const errorMessage = `There has been an error while generating the embed for the
                 current blog. Please try with another theme and contact our Support Team
-                and provide this error: "${eventData.error}"`;
+                and provide this error: "${eventData.error}". The embed generation has fallen
+                back to the Default theme for the current blog.`;
```
eos87, line 91: nit:
```suggestion
                back to the Default SEO Theme for the current blog.`;
```

server/liveblog/blogs/tasks.py
eos87: also, maybe we could extract this into separate function like `def fallback_to_default_theme_if_needed(self, ...` or something similar.

eos87 (changes requested):

eos87 (approved):

PR #1416, opened 2024-10-09
[LBSD-2882] - Create custom voting endpoint for polls

BrianMwangi21:
### Purpose

This PR implements a custom voting endpoint for polls to allow for better poll voting management 

### What has changed

- Added new `client_poll_vote` endpoint to be used to place voted from the embed
- Changed to use of $inc function for incremental voting
- Changed the endpoint called on the embeds
- Added tests for the new endpoint

### Steps to test

2. Open the preview on multiple browsers and place your votes.
3. Observe that the request to send the vote is a new endpoint

Resolves: [LBSD-2882](https://sofab.atlassian.net/browse/LBSD-2882)

[LBSD-2882]: https://sofab.atlassian.net/browse/LBSD-2882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/liveblog/client_modules/client_modules.py
```diff
@@ -557,6 +510,69 @@ def get_blog_posts(blog_id):
...
+            - 400: If the request is invalid or any check fails.
+            - 500: If an error occurs during vote update.
+
+    **TODO:**
+    * Add rate limiter decorator on endpoint
+    """
+    data = request.json
+    option_selected = data.get("option_selected", None)
+    if option_selected is None:
+        return api_error("Error: Option selected is required in request", 400)
```
eos87, line 536: nit:
```suggestion
        return api_error("Please select a voting option."", 400)
```

server/liveblog/client_modules/client_modules.py
```diff
@@ -557,6 +510,69 @@ def get_blog_posts(blog_id):
...
+    poll = polls.find_one(req=None, _id=poll_id)
+    if poll is None:
+        return api_error("Error: Poll not found", 400)
+
+    poll_body = poll.get("poll_body", {})
+    answers = poll_body.get("answers", [])
+    option_exists = any(answer.get("option") == option_selected for answer in answers)
+    if not option_exists:
+        return api_error(
+            f"Error: Option '{option_selected}' not found in poll answers", 400
```
eos87, line 548: also here, 404 as "not found"
```suggestion
            f"Error: Option '{option_selected}' not found in poll answers", 404
```

server/liveblog/client_modules/client_modules.py
```diff
@@ -557,6 +510,69 @@ def get_blog_posts(blog_id):
...
+    """
+    data = request.json
+    option_selected = data.get("option_selected", None)
+    if option_selected is None:
+        return api_error("Error: Option selected is required in request", 400)
+
+    polls = get_resource_service("polls")
+    poll = polls.find_one(req=None, _id=poll_id)
+    if poll is None:
+        return api_error("Error: Poll not found", 400)
```
eos87, line 541: 404 would fit better here I think
```suggestion
        return api_error("Error: Poll not found", 404)
```

server/liveblog/client_modules/client_modules.py
```diff
@@ -557,6 +510,69 @@ def get_blog_posts(blog_id):
...
+        )
+
+    result = polls.find_and_modify(
+        query={"_id": poll_id, "poll_body.answers.option": option_selected},
+        update={"$inc": {"poll_body.answers.$.votes": 1}},
+        new=True,
+    )
+
+    if result is None:
+        return api_error("Error: Unable to update poll votes", 500)
```
eos87, line 558:
```suggestion
        return api_error("Error: Unable to update poll votes", 400)
```

I feel like the 500 here is a bit excessive. Normally 500 is used for catastrophic or unexpected issues. I would suggest to use a 400 as "bad request" or maybe a 422 as "unprocessable entity"

server/liveblog/client_modules/client_modules.py
```diff
@@ -557,6 +510,69 @@ def get_blog_posts(blog_id):
...
+        {"blog": blog_id, "particular_type": "post"}
+    ):
+        for assoc in post_utils.get_associations(post):
+            if assoc.get("residRef") == poll_id:
+                updated_post = post.copy()
+                updated_post["content_updated_date"] = utcnow()
+                get_resource_service("posts").update(
+                    post.get("_id"), updated_post, post
+                )
+                app.blog_cache.invalidate(blog_id)
```
eos87, line 571: following the single responsibility principle (I think you mentioned the other day), I'd suggest to put this into the posts service or in some util method instead, this way we avoid more tight coupling between these. What do you think?

eos87 (changes requested):
Hi @BrianMwangi21, I left some comments, please take a look.

server/liveblog/client_modules/client_modules.py
BrianMwangi21: Sure. No problem

server/liveblog/client_modules/client_modules.py
```diff
@@ -557,6 +510,59 @@ def get_blog_posts(blog_id):
...
+    1. Validate the request data
+    2. Retrieve the poll and validate selected option
+    3. Increment the vote count
+    4. Update associated blog posts to reflect change on embed
+
+    Returns:
+        JSON response with code:
+            - 201: If the vote is successfully recorded.
+            - 400: If the request is invalid or any check fails.
+            - 500: If an error occurs during vote update.
```
eos87, line 528:
```suggestion
            - 422: If an error occurs during vote update.
```

server/liveblog/client_modules/client_modules.py
```diff
@@ -557,6 +510,59 @@ def get_blog_posts(blog_id):
...
+        return api_error("Please select a voting option.", 400)
+
+    polls = get_resource_service("polls")
+    poll = polls.find_one(req=None, _id=poll_id)
+    if poll is None:
+        return api_error("Error: Poll not found", 404)
+
+    poll_body = poll.get("poll_body", {})
+    answers = poll_body.get("answers", [])
+    option_exists = any(answer.get("option") == option_selected for answer in answers)
```
eos87, line 545: this piece got into my mind. Depending on the number of the answers coming from the frontend, this loop could be kinda tricky but we'll leave it like this for now. Once we have the rate limiter it should not be a problem anyways.

eos87 (approved):

eos87 (changes requested):
I just saw the tests are broken.

eos87 (approved):

PR #1418, opened 2024-11-01
[LBSD-2888] - Resolve embed error generation

BrianMwangi21:
### Purpose

This PR fixes the issue with generating the embed for simple theme in relation to the missing `template-embed-utils` file. It also fixes the theme changes applying on associated blogs.

### What has changed

- Fixed generation of simple theme.
- Fixed application of theme changes to associated blogs.

### Steps to test generation of simple theme.

1. Open the Editor and create blog using default theme.
2. Add a few posts.
3. Open the live preview and see the embed in the default theme.
4. Go to the blog settings and change the theme to simple theme. 
5. Observe that when you go back to the live preview the theme has changed.
6. Additionally, go to theme manager and make changes to the theme settings.
7. Observe that the theme changes reflect on the associated blogs.

Resolves: [LBSD-2888](https://sofab.atlassian.net/browse/LBSD-2888)

[LBSD-2888]: https://sofab.atlassian.net/browse/LBSD-2888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/liveblog/themes/themes_assets/amp/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "liveblog-amp-theme",
-  "version": "3.4.35",
+  "version": "3.5.1",
```
eos87, line 3: let's bump it to 3.90.0 so it matches LB version too like Default Theme

server/liveblog/themes/themes.py
```diff
@@ -600,24 +601,26 @@ def _get_publishing_step(self, theme):
                 return STEPS.get(key)
         return STEPS.get("default")
 
-    def _schedule_items(self, items, step, is_blog):
-        """Schedules publishing for given items."""
+    def _schedule_items(self, items, step, is_blog, batch_size=10, delay=1):
+        """Schedules publishing for given items in manageable batches."""
```
eos87, line 605: can you add more context about why we use batches here and what's the logic behind it?

server/liveblog/themes/themes.py
```diff
@@ -600,24 +601,26 @@ def _get_publishing_step(self, theme):
...
         from liveblog.blogs.tasks import publish_blog_embed_on_s3
 
         countdown = 1
+        total_batches = ceil(items.count() / batch_size)
```
eos87, line 609: what is the type of items, is it a cursor? if so, don't you get a deprecation warning?

server/liveblog/themes/themes_assets/simple/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "liveblog-simple-theme",
-  "version": "0.3.6",
+  "version": "0.3.9",
```
eos87, line 3: also here, let's set 3.90.0 too

server/liveblog/themes/themes.py
BrianMwangi21: No, I'm not getting a deprecation warning on this

BrianMwangi21: I've refactored this to use list instead.

server/liveblog/themes/themes_assets/default/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "liveblog-default-theme",
-  "version": "3.9.0",
+  "version": "3.9.0-dev.7",
```
eos87, line 3: I think this should be
```suggestion
  "version": "3.90.0-dev.7",
```

server/liveblog/themes/themes.py
```diff
@@ -600,24 +601,45 @@ def _get_publishing_step(self, theme):
...
+        total_batches = ceil(len(items_list) / batch_size)
+
+        for batch_num in range(total_batches):
+            batch_items = items_list[
+                batch_num * batch_size : (batch_num + 1) * batch_size
+            ]
+
+            for item in batch_items:
+                args = [item.get("_id")] if is_blog else [item.get("blog")]
+                kwargs = {"output": item} if not is_blog else {}
```
eos87, line 635: Something like this would make it easier to read. What do you think?
```suggestion
                if is_blog:
                    args = [item.get("_id")]
                    kwargs = {}
                else:
                    args = [item.get("blog")]
                    kwargs = {"output": item}
```

eos87 (approved):
Looks good, I just left two final comments. When you adjust the version of default theme to 3.90.0 remember to update the references in the child themes + the npm package. Thanks!

client/app/scripts/liveblog-themes/module.js
```diff
@@ -131,12 +131,7 @@ import listTpl from 'scripts/liveblog-themes/views/list.ng1';
                 }
 
                 themes.forEach((theme) => {
-                    // create criteria to load blogs with the theme.
-                    const criteria = {
-                        where: JSON.stringify({'blog_preferences.theme': theme.name}),
-                    };
-
-                    api.blogs.query(criteria).then((data) => {
+                    api.get(`/count_theme_blogs/${theme.name}`).then((data) => {
```
eos87, line 134: We can avoid the extra requests/hits by attaching this info in the "on_fetched" method from the blog service. Like,  

```python
theme["blog_count"] = blog_service.get_from_mongo(...
```

BrianMwangi21: Oh yeah, that makes more sense.

eos87 (approved):

PR #1420, opened 2025-02-07
Remove usages of `notify.pop` as it no longer exists

eos87:
### Purpose
Fix broken references to removed method from notification module.

### Todo
We should double check all usages of `notify` module across the app to see if the little messages are showing/hiding properly


BrianMwangi21 (approved):
LGTM!

PR #1424, opened 2025-02-27
[LBSD-2904] - Investigate and fix root cause of broken references since the 3.90 release [WIP]

BrianMwangi21:
### Purpose

The purpose of this PR is investigate and fix the root cause of broken references since the 3.90 release. 

The primary change in 3.90 in terms of posts being deleted is that the groups were reset to empty arrays. Previously, the post would only be marked as deleted. This had a cascading ripple effect where there was code left behind that still checked for the items in the `groups` object and hence would fail.

### What has changed

- Completely delete the post as well as the items it references.
- Update existing items when a post is updated instead of creating a new one. Many duplicates get created as a result of this and can lead to waste in the database.

**NOTE: New changes might break other things so needs alot of testing**

### Steps to test
Will be listed when PR is fully ready.

Resolves: [LBSD-2904](https://sofab.atlassian.net/browse/LBSD-2904)

[LBSD-2904]: https://sofab.atlassian.net/browse/LBSD-2904?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BrianMwangi21:
Closing this as this is handled in [Release/3.90.4](https://github.com/liveblog/liveblog/pull/1423)

PR #1427, opened 2025-04-29
[LBSD-2909] Strip html tags from LiveblogPostingSchema's `articleBody`

eos87:
### Purpose
To remove the html tags from the schema's article body. At the moment html tags are being rendered and according to the specification, the articleBody should be plain text. 

### What has changed
- Added a small util function to remove the html tags but preserve line breaks (replace `<br>` tags with `\n`)

### To test
- Create a blog with a few posts
- Go to the embed view
- Observe in the `LiveBlogPosting > liveBlogUpdate` that the attribute `articleBody` does not contain html tags

Thanks for checking!

Resolves: [LBSD-2909]

[LBSD-2909]: https://sofab.atlassian.net/browse/LBSD-2909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/liveblog/blogposting_schema/utils.py
+++ b/server/liveblog/blogposting_schema/utils.py
@@ -1,5 +1,7 @@
 import json
 import logging
+from bs4 import BeautifulSoup
+
 from liveblog.posts.utils import (
     get_first_item,
     get_first_item_of_type,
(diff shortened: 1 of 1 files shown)


eos87:
Please ignore the GitHub actions not working on this PR, they are being fixed in #1423

BrianMwangi21 (approved):
Tested it out locally. Looks good to me.

PR #1428, opened 2025-05-19
Fix Instance Settings Persistence

BrianMwangi21:
### Purpose

This PR ensures that instance settings initialized via the `app:initialize_data` command do not overwrite any user-defined changes. Previously, running the command after a user had made updates (e.g. changing limits from 1 to 5) would reset those values back to defaults (like 1). Since restarts or deployments can trigger this command automatically, adding this ensures user configurations don't get lost.

### What has changed

- Updated the `create()` method in the `InstanceSettingsService` to merge incoming settings with existing ones instead of blindly replacing them.
- Implemented a `deep_merge()` function that only adds missing keys or merges nested structures, while leaving user-defined scalar values untouched.
- Set `do_patch = False` for this resource to prevent automatic patching that could unintentionally modify stored settings.
- Updated the `get()` method to explicitly use `get_from_mongo()` so that the latest settings are fetched directly from MongoDB, bypassing Elasticsearch. This ensures that the returned config reflects real-time updates, especially important after changes or restarts.

### Steps to test

1. Start with a clean instance (or drop the instance_settings table if it exists) and run the `app:initialize_data` command. 
2. Go to the Instance Settings page and change a value in the instance settings like setting solo blogs from 1 to 5. Save changes.
3. Run the `app:initialize_data` command again.
4. Confirm that the value remains at 5 and is **not reset** to the default.
5. Make another change and observe that the new changes are reflected.


server/liveblog/instance_settings/instance_settings.py
```diff
@@ -59,6 +60,13 @@ def is_user_authorized(self):
 
         return flask.g.user.get("is_support", False)
 
+    def get(self, req, lookup):
+        """
+        Retrieve instance settings directly from MongoDB, bypassing the Elasticsearch.
+        This ensures that the most up-to-date configuration is returned to the client.
+        """
+        return self.get_from_mongo(req, lookup)
```
eos87, line 68: we probably don't need this. Instead, we should disable elastic search for this resource.

eos87 (commented):
Can you add some tests please? I think it would be good to make sure that the settings do not get overwritten if the initialize command is run. Also, it would be good to check if the settings saved from the UI have higher priority and overwrite those that came from the initialize command.

eos87:
For the tests, behave or unit tests would do the trick (or both 😅). It's up to you.

server/settings.py
```diff
@@ -383,7 +383,7 @@ def env(variable, fallback_value=None):
 SUBSCRIPTION_LEVEL_SOLO = "solo"
 SUBSCRIPTION_LEVEL_TEAM = "team"
 SUBSCRIPTION_LEVEL_NETWORK = "network"
-SUBSCRIPTION_LEVEL = env("SUBSCRIPTION_LEVEL", SUBSCRIPTION_LEVEL_NETWORK)
+SUBSCRIPTION_LEVEL = env("SUBSCRIPTION_LEVEL", SUBSCRIPTION_LEVEL_GO)
```
eos87, line 386: If we're ready to put this in a test instance we should revert this back. I'll ask systems engineers to put it on the go instance to play with it.

PR #1436, opened 2025-07-08
[LBSD-2919] Fix posts not rendered after load more button is hit

eos87:
### Purpose
This fixes the issue of posts not rendered after hitting load more button. The problem was that the frontend was not calculating properly that there were still posts to be fetched/rendered from the backend

### Steps to test
- Create a new blog 
- Create approximately 3 pages of posts (10 posts per page, 28 posts does the trick)
- Switch to embed view and scroll down
- Notice that all the posts are fetched and rendered 
- Notice that when the last post if rendered the `Load More` button is not longer visible
- Repeat the process in different sorting modes (Editorial, Newest first, oldest first) 
- Check if filtering by tags work as expected and render the right number of posts as expected
- Try to break it or discover any other edge case 

Resolves: [LBSD-2919]


BrianMwangi21 (approved):
LGTM!

